PR #20, opened 2019-08-16
set ga option and use vuepress google Analytics plugin

khoa0319:
## Summary
* Add ga default config of vuepress.
* Use vuepress plugin for google Analytics.

## Asana
https://app.asana.com/0/1125458388096768/1135773918803106/f

### DIFF
--- a/packages/dbml-homepage/docs/.vuepress/config.js
+++ b/packages/dbml-homepage/docs/.vuepress/config.js
@@ -18,5 +18,14 @@ module.exports = {
         ''
       ],
     }
-  }
+  },
+  ga: `${require('./secret').ga}`,
+  plugins: [
+    [
+      '@vuepress/google-analytics',
+      {
+        'ga': `${require('./secret').ga}`
+      }
+    ]
+  ]
 }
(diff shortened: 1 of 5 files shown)


packages/dbml-homepage/docs/.vuepress/config.js
```diff
@@ -18,5 +18,14 @@ module.exports = {
...
       ],
     }
-  }
+  },
+  ga: `${require('./secret').ga}`,
+  plugins: [
+    [
+      '@vuepress/google-analytics',
+      {
+        'ga': `${require('./secret').ga}`
```
nntai: should we put Google Analytics ID here instead of separating to a different file?

packages/dbml-homepage/docs/.vuepress/config.js
```diff
@@ -18,5 +18,14 @@ module.exports = {
         ''
       ],
     }
-  }
+  },
+  ga: `${require('./secret').ga}`,
```
nntai: I think `ga` only set in vuepress plugin config

packages/dbml-homepage/docs/.vuepress/config.js
khoa0319: I think Google Analytics ID is kind of a secret key so by convention i think we should not hard-coded it.

packages/dbml-homepage/docs/.vuepress/config.js
khoa0319: My bad, i should have removed it first.

Issue #7, opened 2019-07-10
Support for ON DELETE CASCADE


bf:
It would be nice to have support for `ON DELETE CASCADE` in dbdiagram.io

nntai:
Hi @bf, thank you for your idea. Could you share more about what you need with `ON DELETE CASCADE`?
By this time, if you need this feature for documentation, you could use our note feature to show it in the tooltip when hovering on the table.
```
Table orders {
  id int [pk]
  user_id int [not null, unique]
  status orders_status
  created_at varchar [note: "When order created"]
}

Table order_items {
  order_id int [note: "Will be delete when order deleted"]
  product_id int    
  quantity int [default: 1]
}
```
![Screen Shot 2019-07-12 at 17 09 30](https://user-images.githubusercontent.com/15927583/61120743-d9246900-a4c7-11e9-88fb-27f6a16b5cfc.png)

ben-eb:
@nntai I think there is currently no way to express the below in DBML:

> ```sql
> CREATE TABLE products (
>     product_no integer PRIMARY KEY,
>     name text,
>     price numeric
> );
> 
> CREATE TABLE orders (
>     order_id integer PRIMARY KEY,
>     shipping_address text
> );
> 
> CREATE TABLE order_items (
>     product_no integer REFERENCES products ON DELETE RESTRICT,
>     order_id integer REFERENCES orders ON DELETE CASCADE,
>     quantity integer,
>     PRIMARY KEY (product_no, order_id)
> );
> ```
> Restricting and cascading deletes are the two most common options. `RESTRICT` prevents deletion of a referenced row. `NO ACTION` means that if any referencing rows still exist when the constraint is checked, an error is raised; this is the default behavior if you do not specify anything. (The essential difference between these two choices is that `NO ACTION` allows the check to be deferred until later in the transaction, whereas `RESTRICT` does not.) `CASCADE` specifies that when a referenced row is deleted, row(s) referencing it should be automatically deleted as well.

source: https://www.postgresql.org/docs/11/ddl-constraints.html

ufhy:
up

nntai:
Hi @bf, @ufhy, @ben-eb, we are updating the parser to support syntax: `ON DELETE/ON UPDATE...` but we can not find the best way to express it in DBML syntax. Do you guys have any ideas?
Btw, we did open-source our parser and also the CLI in [packages folder](https://github.com/holistics/dbml/tree/master/packages), you could check it.

[...]

PR #29, opened 2019-09-09
Feature: Referential actions

flemeur:
## Summary
* Add support for ON UPDATE and ON DELETE on foreign keys. In the MySQL documentation this is known as "referential actions"

An example of the syntax:
```
Ref:"orders"."id" < "order_items"."order_id" [update: no action, delete: cascade]
```

This is a work in progress but i would like to get feedback as early as possible. One thing that is still not clear is how the syntax should be for inline references.

## Issue
https://github.com/holistics/dbml/issues/7

* Modified dbml parser
* Modified mysql parser
* Modified json exporter
* Modified mysql exporter
* Modified postgres exporter
* Modified postgres parser
* Modify schemarb parser (not yet implemented)

## Checklist
- [ ] Code Review


nvquanghuy:
Thanks @flemeur for the contribution! From the technical side the code looks okay, the main discussion point should be on the syntax side.

One of the things we try to do at DBML is to be language-agnostic. I'd have to look into how this concept works for other database (PostgreSQL, SQLServer, etc).

One idea we're tinkering is the concept of a "database dialect". So I'd be more comfortable for your PR to go through if dialect is implemented.

```
@mysql
Table orders [storage_type: myisam] {
  id integer
  ...
}
...

Ref: "orders"."id" < "order_items"."order_id" [update: no action, delete: cascade]
```

flemeur:
Hi @nvquanghuy 

Glad to contribute! :+1: 

I agree that we should discus the syntax before moving forward.
I've looked at MySQL, PostgreSQL, sqlite and SQLServer and it looks like they all use the same syntax (apart from how the identifiers are escaped)

* MySQL: https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html#foreign-keys-adding
* PostgreSQL: https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK
* SQLServer: https://docs.microsoft.com/en-us/sql/relational-databases/tables/create-foreign-key-relationships?view=sql-server-2017#create-a-foreign-key-in-an-existing-table
* sqlite: https://sqlite.org/foreignkeys.html#fk_actions

Are there any other databases that you plan to support in the future?

What are your thoughts on the syntax for inline refs? Personally I prefer to define foreign keys inline, if possible, so everything is contained within the same statement.

Currently it looks like this
```
Table posts {
    id integer [primary key]
    user_id integer [ref: > users.id]
}
```
I'm guessing that using a nested pair of square brackets is maybe not what you want :smile: 

I like the idea of having a dialect annotation for things that apply only to a specific DB vendor, fx engine/storage_type. I'm not sure if it applies to foreign key actions though, since they exist for all the vendors I know of.

Anyway, let me know what you think

nvquanghuy:
Thanks @flemeur for the detailed note! Give me a few days to digest this and get back to you properly.

nvquanghuy:
Thanks for your patience, @flemeur.

Regarding syntax standards, I've looked at all the manual docs:
* [PostgreSQL](https://www.postgresql.org/docs/11/sql-createtable.html)
* [MySQL](https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html#foreign-keys-adding)
* [SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-2017)
* [SQLite](https://sqlite.org/foreignkeys.html#fk_actions)
* [SQL:92 text](https://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt)

All of them follows the following (with the exception of `RESTRICT` where SQL:92 and SQLServer doesn't state.

```
[ON DELETE reference_option]
[ON UPDATE reference_option]

reference_option:
  RESTRICT | CASCADE | SET NULL | NO ACTION | SET DEFAULT
```

So I think it's safe to support the 5 values above for delete and update without going through the dialect annotation.

```
Table orders {
  id integer [pk]
}
Table order_items {
  id integer [pk]
  order_id integer [ref: > orders.id]
}
// this
Ref: orders.id < order_items.order_id [update: no action, delete: cascade]
// or this
Ref a {
  orders.id < order_items.order_id
  [update: no action, delete: cascade]
}
```

Regarding syntax for inline refs, as you suggested I don't think nested square brackets make sense at the moment. I'm inclined to say that if you want to define additional constraints like that then it should go through a formal definition without inline definition :)

That means your original proposal is good to go, feel free to proceed with the rest of the PR 👍

flemeur:
Hi again @nvquanghuy 

That is great to hear! I'll continue working on the remaining parts of the PR then. Will ping you when I have something ready for review.

flemeur:
@nvquanghuy I think this is more or less ready for a review.

I haven't changed anything in the schemarb parser, since I don't have any experience with that and I'm not sure how relevant it is. Let me know if this is critical.

packages/dbml-core/src/parse/mysql/parser.pegjs
```diff
@@ -123,27 +123,42 @@ PKSyntax = _ primary_key _ "(" _ names:ListOfNames _ ")"
 {return names}
 
 // FKSyntax: Support "FOREIGN KEY (field[, field]*) REFERENCES `table`(field[, field]*)"
-FKSyntax = _ ("CONSTRAINT"i _ name)? _ foreign_key _ 
+FKSyntax = _ constraint:("CONSTRAINT"i _ name)? _ foreign_key _ 
 	"(" _ fields:ListOfNames _ ")" _
 	references _ table2:table_name _ "(" _ fields2:ListOfNames _ ")" _
-	("ON"i _ ("DELETE"i/"UPDATE"i) _ references_options)?
+	onDelete:FKOnDelete? _
+	onUpdate:FKOnUpdate?
```
phuongduyphan: I've checked with MySQL and Postgres that the order of `ON DELETE action` and `ON UPDATE action` doesn't matter. But if you parsed like above, `ON DELETE action` must always be placed before `ON UPDATE action`. Pls help fix this.

phuongduyphan:
Seems like we have problem with parsing **inline** `FOREIGN KEY ` constraint in both MySQL and Postgres. Ex with MySQL: 
```
CREATE TABLE `users` (
  `id` int PRIMARY KEY AUTO_INCREMENT,
  `name` varchar(255),
  ...
);

CREATE TABLE `orders` (
  `id` int PRIMARY KEY AUTO_INCREMENT,
  `user_id` int NOT NULL,
  FOREIGN KEY (`user_id`) REFERENCES `users` (`id`) ON DELETE CASCADE
);
```
The parsed result object doesn't have attribute `onDelete`.

The problem is that in the parsing process we process the inline refs separately in the `TableBody` block and after that put all inline_refs into refs in `TableSyntax`. However, we haven't assigned `onDelete` or `onUpdate` to the inline_ref so in `TableSyntax` so that the inline_ref doesn't have those attributes

packages/dbml-core/src/parse/mysql/parser.pegjs
flemeur: @phuongduyphan This should be fixed now.

flemeur:
@phuongduyphan Thanks for taking a look at the PR :+1: I'll add some test cases for inline foreign keys and see what i can do to make them work as well

flemeur:
Hi @phuongduyphan I believe i have gotten inline refs to work for the mysql parser so far. Could you take a look and see if I have missed something? I'll see if I can do the same for the postgres parser, but it is a lot more complicated to change so might need some help with that. Thanks in advance!

phuongduyphan:
Hi @flemeur , 

Thanks for your great work again !

I've already added implementation for parsing inline ref with actions for postgres. Can you finally check it again ? If everything's ok, I think the PR is ready to be merged in master.

flemeur:
Hi @phuongduyphan 
Thanks for doing the postgres parser changes. It looks great! I think we're ready to merge the PR now :+1:

PR #90, opened 2020-05-12
Support composite foreign keys

wingombia:
## Summary
- Add support for composite foreign key (import, export, parsing).
- An example of the new composite foreign keys syntax for dbml:  
products.(id, name) < order_items.(product_id, product_name).
- Still support the old syntax: products.id < order_items.product_id.

- Change all parser.pegjs files and exporters to support composite foreign key (in ref section). 
- Endpoint normalized form's fieldId is replaced by fieldIds (array).
- Endpoint's fieldName, field is replace by fieldNames, fields (array).
- Change Endpoint#equals.
- Add error handling for when two endpoints have different number of fields in Ref object.
- Add buildFieldName to export/utils for building a field names string when exporting.
- Change (mostly in test files that have foreign keys definition) and add (mostly in referential_actions test files) new test cases .

## Checklist

- [ ] Code Review


packages/dbml-core/src/parse/dbml/parser.pegjs
```diff
@@ -183,6 +183,21 @@ ref_body
...
+  = field:(RefSingleField/RefMultipleFields) { 
+    if (typeof field === "string") field = [field];
+    return field; 
+  }
+
+RefSingleField
+  =  field:name { return field; }
+ 
+RefMultipleFields
+  = "(" sp* first:RefSingleField rest:(Comma sp* RefSingleField)* ")"  {
```
phuongduyphan, line 197: This input will fail.
`users.(id , name  ) < orders.(  user_id , username  )`

Suggest adding `sp*` before Comma and after `rest` here.

phuongduyphan: Also, pls change `fieldName` to `fieldNames`

packages/dbml-core/src/model_structure/ref.js
```diff
@@ -41,6 +41,11 @@ class Ref extends Element {
     if (this.endpoints[0].equals(this.endpoints[1])) {
       this.error('Two endpoints are the same');
     }
+
+    if (this.endpoints[0].fields.length !== this.endpoints[1].fields.length){
```
phuongduyphan, line 45: Pls fix eslint error here: `Missing space before opening brace`

packages/dbml-core/src/model_structure/endpoint.js
```diff
@@ -9,7 +9,8 @@ class Endpoint extends Element {
 
     this.schemaName = schemaName;
     this.tableName = tableName;
-    this.fieldName = fieldName;
+    this.fieldName = fieldName; // can be an array of names
```
phuongduyphan, line 12: As I see at all the places in this PR, this is always an array of names. 

Suggest changing `fieldName` to `fieldNames` instead.

packages/dbml-core/src/model_structure/endpoint.js
```diff
@@ -55,13 +63,22 @@ class Endpoint extends Element {
       relation: this.relation,
     };
   }
+  
+  setFields (fieldNames, table) {
+    if (typeof fieldNames === "string") fieldNames = [fieldNames]
```
phuongduyphan, line 68: `fieldNames` are always array as the output from pegjs parser so the checking here is not necessary anymore.

packages/dbml-core/src/model_structure/endpoint.js
```diff
@@ -22,28 +23,35 @@ class Endpoint extends Element {
...
+    this.setFields(fieldName, table) 
   }
 
   generateId () {
     this.id = this.dbState.generateId('endpointId');
   }
 
   equals (endpoint) {
-    return this.field.id === endpoint.field.id;
+    if (this.fields.length != endpoint.fields.length) return false;
```
phuongduyphan, line 34: Suggest using strict comparison operator `!==` instead of `!=`

packages/dbml-core/src/model_structure/endpoint.js
```diff
@@ -22,28 +23,35 @@ class Endpoint extends Element {
...
+    return this.fields.length == 1 ? this.field.id === endpoint.field.id : this.compareFields(endpoint);
```
phuongduyphan, line 35: same here, `===` instead of `==`

packages/dbml-core/src/model_structure/endpoint.js
```diff
@@ -55,13 +63,22 @@ class Endpoint extends Element {
...
+  }
 
   setField (field, table) {
     if (!field) {
       this.error(`Can't find field ${shouldPrintSchema(table.schema)
-        ? `"${table.schema.name}".` : ''}"${this.fieldName}" in table "${this.tableName}"`);
+        ? `"${table.schema.name}".` : ''}"${field.name}" in table "${this.tableName}"`);
     }
-    this.field = field;
+    if (!this.field) this.field = field;
```
phuongduyphan, line 80: Why we need to set `this.field = field` here?

packages/dbml-core/src/model_structure/endpoint.js
phuongduyphan: Why we have to divide into 2 cases for comparing fields here? Can `compareFields` compare fields with length 1?

wingombia:
> As I see at all the places in this PR, this is always an array of names.
Suggest changing fieldName to fieldNames instead.

> Why we have to divide into 2 cases in order to compare fields here? Can `compareFields` compare fields with length 1?

> Why we need to set this.field = field here?

oops, I forgot to push the latest commits where I fixed these.

packages/dbml-core/src/parse/dbml/parser.pegjs
```diff
@@ -183,6 +183,21 @@ ref_body
       settings: ref_settings
     };
   }
+//CHANGE
+RefField
+  = sp* field:(RefSingleField/RefMultipleFields) {
```
phuongduyphan, line 188: We didn't allow spaces before `.` so to make it consistent pls remove `sp*` here.

PR #102, opened 2020-06-05
Fix endpoint error handling and support for implicit foreign key reference for mssql

phanbaominh:
## Summary
- Fix error handling when endpoint have fields that are not in endpoint's table.
- Add support for implicit foreign key reference for mssql, only when columns on both sides have the same name, example:
`[user_id] int CONSTRAINT fk_order_user FOREIGN KEY REFERENCES [users]`
## Issue
(issue link here)

- Remove  setField in endpoint and move it to setFields, change field.name to fieldName in error message.
## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/__tests__/parser/mssql-parse/input/inline_referential_actions.in.sql
+++ b/packages/dbml-core/__tests__/parser/mssql-parse/input/inline_referential_actions.in.sql
@@ -1,7 +1,7 @@
 CREATE TABLE [orders] (
   [id] int PRIMARY KEY IDENTITY(1, 1),
   [user_id] int UNIQUE NOT NULL 
-    CONSTRAINT fk_order_user FOREIGN KEY REFERENCES [users] ([id]) 
+    CONSTRAINT fk_order_user FOREIGN KEY REFERENCES [users] 
     ON DELETE NO ACTION,
   [status] orders_status_enum,
   [created_at] varchar(255)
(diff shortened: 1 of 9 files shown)


packages/dbml-core/src/model_structure/endpoint.js
```diff
@@ -71,19 +71,16 @@ class Endpoint extends Element {
   setFields (fieldNames, table) {
     fieldNames.forEach(fieldName => {
       const field = table.findField(fieldName);
-      this.setField(field, table);
+      if (!field) {
+        this.error(`Can't find field ${shouldPrintSchema(table.schema)
+          ? `"${table.schema.name}".` : ''}"${fieldName}" in table "${this.tableName}"`);
+      } else {
+        this.fields.push(field);
```
khanh-tran-quoc, line 78: No need to use `else` here, the process will stop after calling `this.error`

packages/dbml-core/__tests__/parser/mssql-parse/input/inline_referential_actions.in.sql
```diff
@@ -1,7 +1,7 @@
 CREATE TABLE [orders] (
   [id] int PRIMARY KEY IDENTITY(1, 1),
   [user_id] int UNIQUE NOT NULL 
-    CONSTRAINT fk_order_user FOREIGN KEY REFERENCES [users] ([id]) 
+    CONSTRAINT fk_order_user FOREIGN KEY REFERENCES [users]
```
khanh-tran-quoc, line 4: Can u explain why u need to change these test cases?

phanbaominh: For testing implicit fk reference feature that I also implement in this branch

PR #131, opened 2020-08-25
[Enhancement]support symbol for mysql importer

khoahuynhdev:
## Summary
This is valid in mysql
```sql
create table products (
  id int,
  name varchar(255),
  weight_measurement ENUM('lb', 'kg', '%'),
  primary key(`id`)
) engine=innodb default charset=utf8;
```
From dbdiagram mysql import
Screen Shot 2020-08-25 at 4 40 28 PM

The enhancement
Screen Shot 2020-08-25 at 4 43 13 PM

* Add supported symbol to 'ExpChar'

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/src/parse/mysql/parser.pegjs
+++ b/packages/dbml-core/src/parse/mysql/parser.pegjs
@@ -491,7 +491,7 @@ factor = factors:(character+ _ "(" expression ")"
     / (exprCharNoCommaSpace+ &(_/","/");"/endline");")) / exprChar+ &.) {
     	return _.flattenDeep(factors).join("");
     }   
-exprChar = [\',.a-z0-9_+-\`]i
+exprChar = [\',.a-z0-9_+-\`%]i
     / sp
     / newline
     / tab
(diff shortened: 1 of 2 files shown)


packages/dbml-core/src/parse/mysql/parser.pegjs
```diff
@@ -491,7 +491,7 @@ factor = factors:(character+ _ "(" expression ")"
     / (exprCharNoCommaSpace+ &(_/","/");"/endline");")) / exprChar+ &.) {
     	return _.flattenDeep(factors).join("");
     }   
-exprChar = [\',.a-z0-9_+-\`]i
+exprChar = [\',.a-z0-9_+-\`%]i
```
khanh-tran-quoc, line 494: How about other characters `^?&*...` anh?

khoahuynhdev: I think we just need to cover `%` in this case because `%` is usually used as unit of measurement.
I haven't seen use cases for other characters so until the case arises, we can support them later.

khanh-tran-quoc: ok anh

khanh-tran-quoc (approved):

Issue #146, opened 2020-12-04
[Question] Many-to-many relationships


slanden:
I really like that DBML is a thing! I started using a very similar syntax when trying to speed up my database design process, and then I stumbled across this. Thank you for creating what should have been there from the start!

To the question...

I read the [post on dbdiagram.io](https://community.dbdiagram.io/t/tutorial-many-to-many-relationship/412) about why many-to-many relationships aren't possible. I imagine you've thought long and hard about this. So, why couldn't you use something like the following for a many-to-many reference?

```
table carts {
    products integer [ref: >< products.id]
  }
```

nvquanghuy:
Thanks @slanden. This is something we might reconsider, but our current thinking is that: Since DBML resembles the database structures (hence the ability to export to SQL DDL), we support what a normal database supports.

All online literature I find when it comes to define many-to-many relationship in RDBMS always involve a bridge table (`products_carts`), hence DBML will require an explicit definition of the bridge table.

slanden:
I'm not saying a bridge table is unnecessary, I just think of a bridge table as a specific requirement for a complex optimization pattern for a specific type of database (normalization in relational databases). It has no business meaning.

For example, if I'm designing an online store I know I need to store `carts` and `products`. That's the business requirements. A bridge table is more of an implementation detail. `carts` and `products` will always be needed but the need for a bridge table depends on the way the data is stored.

What intrigues me about DBML is the possibility of envisioning the core models (`products` and `carts`) I need for my online store, without being bogged down by database specific requirements, syntax, patterns, etc. In the end, DBML would give a lean and simple, high-level overview of, "what data models are needed for this project" without losing any real information.

This is all my opinion, but with that perspective on bridge tables I think they could be represented implicitly with the syntax in my previous comment or something similar. The CLI tool would create the necessary DDL statements like any other table, only these would be generated so would have default table and column names.

What types of issues or challenges, do you foresee this presenting?

nvquanghuy:
@slanden apologize for dropping the ball on this and I hope you're still using DBML in some capacity.

Based on your suggestion, along with the discussions in [this thread](https://community.dbdiagram.io/t/tutorial-many-to-many-relationship/412/9), we'll be looking into supporting this.

Please help upvote them here: https://community.dbdiagram.io/t/tutorial-many-to-many-relationship/412/9

PR #262, opened 2022-05-24
Support many to many relationship

baolequoc:
## Summary
* Implement parser for many to many relationship
* Implement export for many to many relationship

## Issue
(https://github.com/holistics/dbml/issues/146#issuecomment-739096912)

* Add syntax `<>` for ref in parser.pegjs
* Add function `getRelations(operator)` to get relation for every operator relation in parser.pegjs
* Add function `buildJunctionFields1` to get fields that will the primary key of table first in PosgresExporter, MySqlExporter, SQLServerExporter
* Add function `buildJunctionFields2` to get fields that will the primary key of table second, if fields exist, it wils generate new name  in PosgresExporter, MySqlExporter, SQLServerExporter
* Add function `buildNewTableName` to create name for new table in PosgresExporter, MySqlExporter, SQLServerExporter
*  Add function `buildTableManyToMany` to generate SQL create new table in PosgresExporter, MySqlExporter, SQLServerExporter
* Add function `buildIndexManytoMany ` to generate SQL create  index for fields of new table in PosgresExporter, MySqlExporter, SQLServerExporter
*  Add function `buildForeignKeyManyToMany` to generate SQL create  foregin key for new table in PosgresExporter, MySqlExporter, SQLServerExporter
* Replace the logic handle export ref in `exportRefs.js` 
```   const refOneIndex = ref.endpointIds.findIndex(endpointId => model.endpoints[endpointId].relation === '1');
      const refEndpointIndex = refOneIndex === -1 ? 0 : refOneIndex;
      ......
      if(refOneIndex === -1) { // case [*,*] (many to many)
      .......
      } else { // case [1, *] (one to many), [1, 1],  (one to one) [*,1] (many to one)
      ........   
      }
```
## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/src/export/MysqlExporter.js
+++ b/packages/dbml-core/src/export/MysqlExporter.js
@@ -124,13 +124,13 @@ class MySQLExporter {
   static buildJunctionFields2 (fieldIds, model, firstTableKey) {
     const mapFieldKeys = new Map();
     fieldIds.map((fieldId) => {
-      let key = `${model.tables[model.fields[fieldId].tableId].name}_${model.fields[fieldId].name}`;
+      let fieldName = `${model.tables[model.fields[fieldId].tableId].name}_${model.fields[fieldId].name}`;
       let count = 1;
-      while (firstTableKey.has(key)) {
-        key = `${model.tables[model.fields[fieldId].tableId].name}_${model.fields[fieldId].name}(${count})`;
+      while (firstTableKey.has(fieldName)) {
+        fieldName = `${model.tables[model.fields[fieldId].tableId].name}_${model.fields[fieldId].name}(${count})`;
         count += 1;
       }
-      mapFieldKeys.set(key, model.fields[fieldId].type.type_name);
+      mapFieldKeys.set(fieldName, model.fields[fieldId].type.type_name);
     });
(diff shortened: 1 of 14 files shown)


packages/dbml-core/src/export/MysqlExporter.js
```diff
@@ -134,29 +188,58 @@ class MySQLExporter {
...
       return line;
     });
 
     return strArr;
   }
 
+  static buildNameNewTable (tableFirst, tableSecond, tables) {
+    let nameNewTable = `${tableFirst}_${tableSecond}`;
+    let count = 1;
+    while (true) {
```
NQPhuc, line 233: This is very inefficient and won't work when there are 2 n-n refs from the same 2 tables (but in different fields). Should instead create a global set of used table names.

packages/dbml-core/src/export/PostgresExporter.js
```diff
@@ -1,4 +1,4 @@
-import _ from 'lodash';
+import _, { keys } from 'lodash';
```
NQPhuc, line 1: Should remove this if not used

NQPhuc (approved):
LGTM!

PR #300, opened 2022-10-26
Add token for note

nguyenalter:
## Summary
* Add token (location) for note
* Update types and test cases

## Issue
(issue link here)

- Update DBML parser to return token from `ObjectNote`
- Update other SQL parsers to return the note `object` instead of `string`
- Update model structure to store `noteToken`
- Update types annotation for model structures

## Checklist

- [ ] Code Review


NQPhuc (approved):
LGTM. However, you should note down changes that only affect code styles so the reviewers don't have to ponder around so much.

Issue #230, opened 2022-01-26
Make duplicated reference error more clear


hustlahusky:
I got an error when trying to convert mysql dump with `sql2dbml`

```
ERROR: You have a syntax error at "dump.sql" line 1 column 1. Reference with same endpoints duplicated
```

This error doesn't really help to figure it out what's wrong with my dump. I found that this error triggered [here](https://github.com/holistics/dbml/blob/aa084e72bd36fe87a033b9ea1de83cc6e333fe2f/packages/dbml-core/src/model_structure/schema.js#L99).

I think this error message can be more clear if you can add reference details like name, endpoints, real position in source file.

Some people on the forums have also encountered this error: <https://community.dbdiagram.io/t/importing-from-mysql-file/762/3>

Eunoia:
I was able to fix this by selectively deleting portions of the file until the error went away. It looks like it had issue with the final alter table statement. Here is the obfuscated statement. 

`ALTER TABLE ONLY public.ducks
    ADD CONSTRAINT ducks_pond_id_fkey_fkey FOREIGN KEY (pond_id_fkey) REFERENCES public.ponds(id);`

nguyenalter:
fixed in #327

PR #316, opened 2022-12-12
Make the reference duplicated error more clear

vunguyen-it:
## Summary
* Make duplicated reference error more clear

## Issue

https://github.com/holistics/dbml/issues/230

no

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/src/model_structure/schema.js
+++ b/packages/dbml-core/src/model_structure/schema.js
@@ -82,7 +82,7 @@ class Schema extends Element {
 
   checkRef (ref) {
     if (this.refs.some(r => r.equals(ref))) {
-      ref.error('Reference with same endpoints duplicated');
+      ref.error(`Reference ${ref.name} has duplicated endpoint`);
     }
   }
 


nguyenalter:
@vunguyen-it your fix only works if the endpoint has a `name` property, otherwise it will show the error `Reference null has duplicated endpoint` which is more ambiguous.

For example, this PostgreSQL code will raise the above error after applying your fix:
  ``` SQL
CREATE TABLE "ecommerce"."users" (
    "id" int PRIMARY KEY,
    "name" varchar,
    "createdAt" timestamp
);

CREATE TABLE "ecommerce"."orders" (
    "id" int PRIMARY KEY,
    "createdBy" int,
    "createdAt" timestamp,
    FOREIGN KEY(createdBy) REFERENCES "ecommerce"."users"(id)
);

ALTER TABLE "ecommerce"."orders" ADD FOREIGN KEY(createdBy) REFERENCES "ecommerce"."users"(id);
  ```
My suggestion: change the error message to `Reference with the same endpoints already exists: "ecommerce"."orders"("createdBy") references "ecommerce"."users"("id")`

nguyenalter:
fixed in #327

PR #328, opened 2023-02-14
Init CI test

nguyenalter:
## Summary
* Init CI test

## Issue
(issue link here)

- Update `package.json`scripts
- Add `test.yml` CI workflow for Github Actions

## Checklist

- [ ] Code Review


.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
+name: Unit tests CI
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
huyphung1602, line 9: Which ubuntu we are using on our production env? I think we could use Ubuntu 20.04 instead.

huyphung1602: In case we are using Ubuntu 22.04, we should use `runs-on: ubuntu-22.04` too. We should stick to a version instead of using Github latest version.

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+      matrix:
+        node-version: [16.x]
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'yarn'
```
huyphung1602, line 21: Should re-trigger the job to learn if the current cache is working correctly nha:
image

huyphung1602:
I think we should start to add the Technical Doc related to a PR directly in the PR description nha. cc @TeaNguyen and @NQPhuc 

Please help add this doc into the PR description nha @nguyenalter: https://www.notion.so/holistics/Technical-Docs-Setup-CI-tests-550a00a71aea452f9614892565930004

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Install dependencies
+      run: yarn bootstrap
```
NQPhuc, line 23: Before running `yarn bootstrap`, please consider running `yarn` first to install the `learna` version specify in `package.json` instead of relying on to `npx` install whatever it want.
ee

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
+name: Unit tests CI
+on:
+  push:
+    branches: [ master ]
+  pull_request:
```
huyphung1602, line 5: > By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

Please help re-check if they are the event activities you want to trigger this job nha. I also raise a question related to this one in the TD.

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+        node-version: [16.x]
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'yarn'
+    - name: Install dependencies
```
huyphung1602, line 22: In case a PR has many commits, each commit will trigger the job. We should cache the result after installing the dependencies to reduce the billable time. It is spending 56 seconds here.

image

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [16.x]
+
+    steps:
+    - uses: actions/checkout@v3
```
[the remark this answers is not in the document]
nguyenalter: According to the [docs](https://github.com/actions/checkout#checkout-v3), the default `fetch-depth` value is `1` -> only fetches the testing commit so I don't think we should change it to `10`

.github/workflows/test.yml
nguyenalter: using `ubuntu-22.04` now

.github/workflows/test.yml
nguyenalter: The cache is available in 2nd job
image

.github/workflows/test.yml
nguyenalter: The running time is reduced due to the yarn cache from `actions/setup-node@v3`
image

.github/workflows/test.yml
nguyenalter: I think we don't need `reopened` event

.github/workflows/test.yml
nguyenalter: resolved

.github/workflows/test.yml
nguyenalter: now only using `opened` and `synchronize` PR to master events and push to master event

huyphung1602 (approved):
LGTM

PR #352, opened 2023-04-07
docs: more reasonable example of default

raphaeleidus:
## Summary
* Existing string example for column default doesn't really make sense to have a default gender.
* Swap it out for a `source` column. This is a common field referring to marketing channels and `'direct'` would be a reasonable default if no other source is provided.

## Issue
none

Just a doc example update

## Checklist

- [X] Documentation (if necessary)

### DIFF
--- a/dbml-homepage/docs/docs/README.md
+++ b/dbml-homepage/docs/docs/README.md
@@ -160,7 +160,7 @@ Example,
         id integer [primary key]
         username varchar(255) [not null, unique]
         full_name varchar(255) [not null]
-        gender varchar(1) [default: 'm']
+        source varchar(255) [default: 'direct']
         created_at timestamp [default: `now()`]
         rating integer [default: 10]
     }


dbml-homepage/docs/docs/README.md
```diff
@@ -160,7 +160,7 @@ Example,
         id integer [primary key]
         username varchar(255) [not null, unique]
         full_name varchar(255) [not null]
-        gender varchar(1) [default: 'm']
+        source varchar(255) [default: 'direct']
```
huyphung1602, line 163: Hi Raphael,

Thank you for your PR.
We believe that the we are trying to add a case of `varchar(1)` in this example. How about my suggestion?

```
gender varchar(1) [not null]
source varchar(255) [default: 'direct']
```

huyphung1602:
Hello Raphael, since there has been no response. To make the example more reasonable, we opened a PR at https://github.com/holistics/dbml/pull/426. I'd like to close this PR right now. Thank you very much.

Feel free to open a new PR to address the new issues the next time.

PR #414, opened 2023-08-01
1. Alt parser

huy-dna:
## Summary
* Add an alternative hand-written dbml parser (not integrated yet).
* How the parser's currently work: https://dbml-explorer-demo-1.vercel.app/
* Next PR: #419 

## Issue
[Asana](https://app.asana.com/0/1125458388096768/1205090003669536/f)

* Add an alternative dbml parser.

## Checklist

- [ ] Code Review


NQPhuc:
Please ensure all CI tests passed

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
```diff
@@ -0,0 +1,567 @@
...
+          this.addToken(SyntaxTokenKind.SEMICOLON);
+          break;
+        case ':':
+          this.advance();
+          this.addToken(SyntaxTokenKind.COLON);
+          break;
+        case '<':
+          this.advance();
+          if (this.peek() !== '=') {
+            this.addToken(SyntaxTokenKind.LT);
```
phuongduyphan, line 153: Instead of classifying for each operator `<=`, `>`, etc., should we use one kind `OPERATOR` for all of the operators? Doing that in the future if we want to support new operators, we don't need to parse again. Also, same operator in different contexts has different names/meaning (i.e. `>` this could be greater than in binary expression, but it could be many-to-one operator in relationship expression)

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
```diff
@@ -0,0 +1,567 @@
...
+      case 'f':
+        char = 'f';
+        this.advance();
+        break;
+      case 'u': {
+        let hex = '';
+        for (let i = 0; i <= 3; i += 1) {
+          if (this.peek() && isAlphaNumeric(this.peek()!)) {
+            hex += this.advance();
+            char = String.fromCharCode(parseInt(hex, 16));
```
phuongduyphan, line 552: Should we put this outside of the loop?

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
```diff
@@ -0,0 +1,567 @@
...
+          this.start - 3,
+          this.current - this.start + 6,
+          string,
+        ),
+      );
+      this.skip(3);
+    }
+  }
+
+  functionExpression() {
```
phuongduyphan, line 335: This method logic is quite similar to `quotedVariable`, `singleLineStringLiteral`, etc... Should we have a generic function for all of them?

packages/dbml-core/src/parse/dbml/src/lib/utils.ts
```diff
@@ -0,0 +1,15 @@
+export function isAlpha(char: string): boolean {
+  const [c] = char;
+
+  return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c === '_';
```
phuongduyphan, line 4: why is `_` included in this check?

phuongduyphan:
This PR is quite large @huy-dna, next time suggest we create multiple smaller PRs (i.e one for lexer, one for parser)

phuongduyphan:
Also pls help me benchmark the new parser with the old pegjs one

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
```diff
@@ -0,0 +1,567 @@
...
+        this.advance();
+        c = this.peek();
+      }
+      this.errors.push(
+        new ParsingError(ParsingErrorCode.INVALID, 'Invalid number', this.start, this.current),
+      );
+    } else {
+      this.addToken(SyntaxTokenKind.NUMERIC_LITERAL);
+    }
+  }
```
phuongduyphan, line 445: Why is this method so complex than it is in AML @huy-dna?

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
```diff
@@ -0,0 +1,567 @@
...
+                ParsingErrorCode.EXPECTED_THINGS,
+                `Unexpected token ${c}`,
+                this.start,
+                this.current - 1,
+              ),
+            );
+            break;
+          }
+      }
+      this.closeGap();
```
phuongduyphan, line 222: What is the meaning of this?

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
```diff
@@ -0,0 +1,567 @@
...
+        triviaList.push(token);
+        if (token.kind === SyntaxTokenKind.NEWLINE) {
+          if (lastNonTrivia) {
+            lastNonTrivia.trailingTrivia = triviaList;
+            startOfLine = true;
+            lastNonTrivia = undefined;
+            triviaList = [];
+          }
+        }
+        continue;
```
phuongduyphan, line 247: this code block has 3 nested if with a continue statement. I think the logic is quite hard to follow and understand what's going on here. It's great if you could add some notes for this, or refactor to a simpler logic if possible

packages/dbml-core/src/parse/dbml/src/lib/result.ts
```diff
@@ -0,0 +1,34 @@
+import { ParsingError } from './errors';
+
+export default class Result<T> {
```
phuongduyphan, line 3: It's better to add some note to explain what are this `Result` class used for? Is it the result of the parsing process or it could be used for other generic use cases?

phuongduyphan: If it could be used for other generic use cases, does it make sense to hardcode the error type for this?

packages/dbml-core/src/parse/dbml/src/lib/result.ts
```diff
@@ -0,0 +1,34 @@
...
+  private value?: T;
+
+  private errors: ParsingError[];
```
phuongduyphan, line 6: If it could be used for other generic use cases, does it make sense to hardcode the error type for this?

packages/dbml-core/src/parse/dbml/src/lib/errors.ts
```diff
@@ -0,0 +1,28 @@
+export enum ParsingErrorCode {
```
phuongduyphan, line 1: Why do we need this?

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
huy-dna: This one also checks for adjacent alphabetical characters after the digits. If there are such characters, the number is marked invalid.
In AML, the lexer doesn't check this, which leads to the problem mentioned in this:
https://app.asana.com/0/1152839153201473/1205071037227754/f

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
```diff
@@ -0,0 +1,567 @@
...
+    );
+  }
+
+  lex(): Result<SyntaxToken[]> {
+    this.init();
+    while (!this.isAtEnd()) {
+      const c: string = this.peek()!;
+      switch (c) {
+        case ' ':
+          this.advance();
```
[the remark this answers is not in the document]
huy-dna: I agree with the `skip()` method. About the repetition of `advance()`, I have thought about it before. However, at the same time, I would like methods such as `identifier`, `functionExpression` to be standalone, independent of the context they were called -- not that it was called in a situation where the first character has been consumed by `advance()`.

packages/dbml-core/src/parse/dbml/src/lib/errors.ts
huy-dna: It's a way to map a type of error to a code. It isn't strictly needed and I only include it if this may be desired in the future.

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
huy-dna: I have added a method called extractString

huy-dna:
> Also pls help me benchmark the new parser with the old pegjs one

I have performed a cursory benchmark for the two:
For the following program:
image

Running it for about ten times, the total time of:
* The pegjs one:
  
  image

* The new one:
  
  image

huy-dna:
Benchmark results:

* 18k
  * Pegjs (5 times):
  
    image

  * New one (10 times):
  
    image

* 25k
  * Pegjs (5 times):
     
   image


  * New one (10 times):
    
    image

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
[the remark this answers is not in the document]
phuongduyphan: > I would like methods such as identifier, functionExpression to be standalone, independent of the context they were called

So why do you prefer this em? Is there any specific reason behind?

packages/dbml-core/src/parse/dbml/src/lib/utils.ts
phuongduyphan: If so I don't think the name of this function reflects that e

packages/dbml-core/src/parse/dbml/src/lib/lexer/lexer.ts
phuongduyphan: Oh I see. But I don't think the lexer should care about it. IMO, this logic should be in the parser. TypeScript is still able to scan this one `2const` as two separate tokens `2` and `const`

packages/dbml-core/src/parse/dbml/src/lib/errors.ts
phuongduyphan: What are the use cases in the future that would need this em?

packages/dbml-core/src/parse/dbml/src/lib/parser/parser.ts
```diff
@@ -0,0 +1,827 @@
...
+    this.contextStack = new ParsingContextStack();
+  }
+
+  private advance(): SyntaxToken {
+    return this.tokens[this.current++];
+  }
+
+  private peek(lookahead: number = 0): SyntaxToken | undefined {
+    if (lookahead + this.current >= this.tokens.length) {
+      return undefined;
```
phuongduyphan, line 64: Instead of returning undefined, could we consider returning the last EOF token instead? By doing this, the return type of our method becomes uniform, and we can eliminate the need for an undefined check every time we invoke it.

packages/dbml-core/src/parse/dbml/src/lib/parser/parser.ts
```diff
@@ -0,0 +1,827 @@
...
+
+  private init() {
+    this.current = 0;
+    this.errors = [];
+    this.invalid = [];
+    this.contextStack = new ParsingContextStack();
+  }
+
+  private advance(): SyntaxToken {
+    return this.tokens[this.current++];
```
phuongduyphan, line 59: Should we handle the case when `this.current` is out of bound?

packages/dbml-core/src/parse/dbml/src/lib/parser/parser.ts
```diff
@@ -0,0 +1,827 @@
...
+} from './nodes';
+
+export default class Parser {
+  private tokens: SyntaxToken[];
+
+  private current: number = 0;
+
+  private errors: ParsingError[] = [];
+
+  private invalid: (SyntaxToken | SyntaxNode)[] = [];
```
phuongduyphan, line 37: What is this `invalid` used for?

packages/dbml-core/src/parse/dbml/src/lib/parser/parser.ts
```diff
@@ -0,0 +1,827 @@
...
+      throw error;
+    }
+  }
+
+  parse(): Result<SyntaxNode & { kind: SyntaxNodeKind.PROGRAM }, ParsingError> {
+    const body: ElementDeclarationNode[] = [];
+
+    this.init();
+
+    while (this.peek() && this.peek()?.kind !== SyntaxTokenKind.EOF) {
```
phuongduyphan, line 113: Can we refactor this line to `while (!this.isAtEnd) {`?

packages/dbml-core/src/parse/dbml/src/lib/parser/parser.ts
```diff
@@ -0,0 +1,827 @@
...
+      try {
+        body.push(this.elementDeclaration());
+      } catch (e) {
+        if (!(e instanceof ParsingError)) {
+          throw e;
+        }
+        if (!(this.peek()?.kind === SyntaxTokenKind.EOF)) {
```
phuongduyphan, line 120: same here, can we use `!this.isAtEnd()`?

packages/dbml-core/src/parse/dbml/src/lib/parser/nodes.ts
```diff
@@ -0,0 +1,575 @@
+import { SyntaxToken } from '../lexer/tokens';
+
+export interface SyntaxNode {
+  kind: SyntaxNodeKind;
+  start: number;
```
phuongduyphan, line 5: Does this start count from node's leading trivia (comments, etc.) or from the node's start itself?

packages/dbml-core/src/parse/dbml/src/lib/parser/nodes.ts
```diff
@@ -0,0 +1,575 @@
...
+  | ListExpressionNode
+  | TupleExpressionNode
+  | CallExpressionNode
+  | PrimaryExpressionNode
+  | FunctionExpressionNode
+  | AccessExpressionNode;
+
+export type ExpressionNode =
+  | LiteralElementExpressionNode
+  | NormalFormExpressionNode
```
phuongduyphan, line 212: This `NormalFormExpressionNode` also contains `LiteralElementExpressionNode` and `FunctionApplicationNode`, so what is the point of creating a new type `ExpressionNode`? :pepe-facepalm

packages/dbml-core/src/parse/dbml/src/lib/parser/nodes.ts
```diff
@@ -0,0 +1,575 @@
...
+        this.end = lastName.offset + lastName.length - 1;
+      }
+    }
+    this.name = name;
+    this.valueOpenColon = valueOpenColon;
+    this.value = value;
+  }
+}
+
+export type NormalFormExpressionNode =
```
phuongduyphan, line 196: What does "normal form expression" mean?

packages/dbml-core/src/parse/dbml/src/lib/parser/parser.ts
```diff
@@ -0,0 +1,829 @@
...
+      this.peek()?.kind !== SyntaxTokenKind.LBRACE
+    ) {
+      try {
+        name = this.normalFormExpression();
+      } catch (e) {
+        if (e instanceof ParsingError) {
+          this.synchronizeElementDeclarationName();
+        } else {
+          throw e;
+        }
```
phuongduyphan, line 156: This logic is duplicated in multiple places, suggest we should have a dedicated function for this

packages/dbml-core/src/parse/dbml/src/lib/parser/parser.ts
```diff
@@ -0,0 +1,829 @@
...
+
+    let attributeList: ListExpressionNode | undefined;
+    if (this.check(SyntaxTokenKind.LBRACKET)) {
+      attributeList = this.listExpression();
+    }
+
+    let body: ExpressionNode | BlockExpressionNode | undefined;
+    let bodyOpenColon: SyntaxToken | undefined;
+
+    if (!this.check(SyntaxTokenKind.COLON, SyntaxTokenKind.LBRACE)) {
```
phuongduyphan, line 182: how about checking in this order em: 
```
if (this.match(SyntaxTokenKind.COLON)) {
  ...
} else if (this.match(SyntaxTokenKind.LBRACE)) {
  ...
} else {
  ...
}
```

IMO, checking in this order is more natural and easier to follow than yours

packages/dbml-core/src/parse/dbml/src/lib/parser/nodes.ts
```diff
@@ -0,0 +1,575 @@
...
+    body: BlockExpressionNode | ExpressionNode;
+  }) {
+    this.start = type.offset;
+    this.end = body.end;
+    this.type = type;
+    this.name = name;
+    this.as = as;
+    this.alias = alias;
+    this.attributeList = attributeList;
+    this.bodyOpenColon = bodyOpenColon;
```
phuongduyphan, line 112: where does `bodyOpenBrace` and `bodyCloseBrace` get set em?

packages/dbml-core/src/parse/dbml/src/lib/errors.ts
huy-dna: This is not very useful at current as it's still rudimentary. You can think of it's like HTTP 404 error code. AML has error codes, too :v.

packages/dbml-core/src/parse/dbml/src/lib/parser/nodes.ts
huy-dna: It counts from the content start, not including the trivia. It defines the start of the yellow highlight in the playground. Should I name it more explicitly?

packages/dbml-core/src/parse/dbml/src/lib/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+  expression: VariableNode & { variable: { kind: SyntaxTokenKind.IDENTIFIER } };
+} {
+  return (
+    value instanceof PrimaryExpressionNode &&
+    value.expression instanceof VariableNode &&
+    value.expression.variable.kind === SyntaxTokenKind.IDENTIFIER
+  );
+}
+
+export function last<T>(array: T[]): T | undefined {
```
phuongduyphan, line 86: One tip: for those utility list functions we could use [lodash](https://lodash.com/) instead of rewriting e

PR #439, opened 2023-09-16
11. Add support for editor

huy-dna:
## Summary
* Improve many parts of the parser to support editor more easily
* Add `go to definition`, `go to references` and ` suggestions` monaco-based services.

## Issue
(issue link here)

## Checklist

- [ ] Code Review


packages/dbml-core/src/parse/dbml/src/compiler.ts
```diff
@@ -1,52 +1,535 @@
...
+          }
+
+          return new Some(id);
+        },
+      ),
+    },
+  };
+
+  // A namespace for parsing-related utility
+  readonly parse = {
```
phuongduyphan, line 253: divide queries into namespaces like this is good. We could apply this to AML

packages/dbml-core/src/parse/dbml/src/compiler.ts
```diff
@@ -1,52 +1,535 @@
...
+    _: this.createQuery(
+      Query._Interpret,
+      (): Report<
+        Readonly<{ ast: ProgramNode; tokens: SyntaxToken[]; rawDb?: Database }>,
+        CompileError
+      > => {
+        const parseRes = new Lexer(this.source)
```
phuongduyphan, line 260: One optimization is to divide this query into smaller sub-queries such as parse and analyze as in some cases of editor support (i.e. go to definition, go to references), running interpreter is unnecessary. This would cater for the case of in the future, we want to run language server in multiple workers and in each worker it only compute for things it needs.

phuongduyphan:
The editor doesn't show suggestions for > 2-level property access expression: 

image

packages/dbml-core/src/parse/dbml/src/services/suggestions.ts
```diff
@@ -0,0 +1,731 @@
...
+        nameStack.unshift(curToken.value);
+      } else {
+        return noSuggestions();
+      }
+      curIter = curIter.back();
+      curToken = curIter.value().unwrap_or(undefined);
+    }
+
+    return {
+      suggestions: this.compiler.membersOfName(nameStack).map(({ kind, name }) => ({
```
phuongduyphan, line 598: Why not doing like this, take this schema property access expression as an example: `v1.users.[suggestions]`:

- Finding the name stack
- Get the node of the name `v1` (call the `.containers(offset)`, get the most inner node for example)
- Get symbol of this node (`node.symbol`)
- Then we resolve the member's chain recursively (v1 -> users -> [suggestions])

instead of guessing and using the current ast's symbol table in `ofName`? In the future, if the symbol of this node is declared in different files, using the current ast's symbol table is no longer correct.

huy-dna:
> The editor doesn't show suggestions for > 2-level property access expression:
> 
> image

I have fixed this but I haven't pushed to dbdiagram frontend yet 🥲

packages/dbml-core/src/parse/dbml/src/compiler.ts
```diff
@@ -1,52 +1,535 @@
...
+            })) :
+          []),
+    ),
+  };
+
+  // Return all possible symbols corresponding to a stack of name
+  membersOfName = this.createQuery(
+    Query.MembersOfName,
+    (
+      nameStack: string[],
```
phuongduyphan, line 375: This implies the name stack should always originate from the outer ast scope. What if we want to find members for name declared in some local scope? 

```
Table users {
  Enum status {
    ok,
    failed
  }

  status.[suggestions]
}
```

packages/dbml-core/src/parse/dbml/src/compiler.ts
```diff
@@ -1,52 +1,535 @@
...
+    ): readonly Readonly<{ symbol: NodeSymbol; kind: SymbolKind; readonly name: string }>[] =>
+      this.symbol.ofName(nameStack).flatMap(({ symbol }) => this.symbol.members(symbol)),
+  );
+
+  // Return information about the enclosing scope at the point of `offset`
+  scope = this.createQuery(
```
phuongduyphan, line 381: Why we don't have `.scope()` for a symbol?

packages/dbml-core/src/parse/dbml/src/compiler.ts
huy-dna: I did want to do that to, however, currently the interpreter would append more errors to `errors`, which means if I split it to another query, the cached `errors` would be mutated if I later call the interpreter after the parser. 🤔 I haven't come up with an appropriate solution. maybe we can split parsing errors and interpreter errors?

packages/dbml-core/src/parse/dbml/src/compiler.ts
huy-dna: Yeah, `membersOfName` should be called something like `publicMembersOfName`. To handle this case, we have the `symbol.members` query though, which is passed the symbol of `status` and use the symbol table in that symbol to do the lookup.

huy-dna:
> The editor doesn't show suggestions for > 2-level property access expression:
> 
> image

I have fixed this but the dbdiagram-frontend demo hasn't been updated with this yet 😸

packages/dbml-core/src/parse/dbml/src/services/definition.ts
```diff
@@ -0,0 +1,43 @@
...
+      if (
+        node?.referee?.declaration &&
+        [SyntaxNodeKind.PRIMARY_EXPRESSION, SyntaxNodeKind.VARIABLE].includes(node?.kind)
+      ) {
+        const { startPos, endPos } = node.referee.declaration;
+
+        return [
+          {
+            range: {
+              startColumn: startPos.column + 1,
```
phuongduyphan, line 30: Why we +1 here em?

packages/dbml-core/src/parse/dbml/src/services/suggestions.ts
```diff
@@ -0,0 +1,731 @@
...
+
+  provideCompletionItems(model: TextModel, position: Position): CompletionList {
+    const offset = getOffsetFromMonacoPosition(model, position);
+
+    let iter = TokenSourceIterator.fromOffset(this.compiler, offset);
+
+    // Return `true` if `iter` is before any other non-trivial tokens
+    // Happens when:
+    // A trivial character has just been typed and it's before every non-trivial ones
+    // or autocompletion is triggered before any non-trivial tokens
```
phuongduyphan, line 48: pls give an example em

PR #459, opened 2023-10-12
Fix old test files of DBML

huy-dna:
## Summary
* Inspect and fix old tests.
* Test input changes: See the commit history
* Test output changes: All output changes fall into one of these:
   * 2 trivial output changes:
     Some fields are reordered when serialized.
     Some extra newlines during note content normalization (removing spaces at the start of a line).
   * 1 major output change: Offset calculation
      * Subfield's offset:
        * The new parser sets the span of a subfield from the start of its callee till the end of its last argument.
        * The old parser always takes the whole line to be a subfield's span (except for the first subfield, it doesn't consider leading spaces at all).
      * Ref endpoint's offset:
         * The new parser sets the span of an endpoint to be the span of the endpoint's expression.
         * The old parser takes the whole ref to be an endpoint's span.

## Issue
(issue link here)

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/__tests__/parser/dbml-parse/input/array_type.in.dbml
+++ b/packages/dbml-core/__tests__/parser/dbml-parse/input/array_type.in.dbml
@@ -1,9 +1,9 @@
 Table sal_emp {
   name text
-  pay_by_quarter int[] [not null]
-  schedule text[][] [null]
+  pay_by_quarter "int[]" [not null]
+  schedule "text[][]" [null]
 }
 
 Table tictactoe {
-  squares integer[3][3]
+  squares "integer[3][3]"
 }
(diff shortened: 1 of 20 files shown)


packages/dbml-core/__tests__/parser/dbml-parse/input/table_element.in.dbml
nguyenalter, line 1: I don't get it. Why do we need to change this test case?

packages/dbml-core/__tests__/parser/dbml-parse/input/default_tables.in.dbml
```diff
@@ -1,7 +1,7 @@
 Table orders {
   id int [pk, default: 123]
   user_id int [not null, unique]
-  status varchar [default: "Completed"]
+  status varchar [default: 'Completed']
```
nguyenalter, line 4: The output must be the same so I don't think we need to change the string quote.

packages/dbml-core/__tests__/parser/dbml-parse/input/array_type.in.dbml
nguyenalter, line 1: This change is not necessary IMO.

nguyenalter (commented):
I think we should avoid changing the testcase inputs if there is no issue.

Issue #450, opened 2023-10-10
Need support comment on mysql and another


tritueviet:
ex: create table schema_migrations
(
    version bigint     not null
        primary key,
    dirty   tinyint(1) not null comment "dirty 1 for dirty other is not dirty"
);

need convert sql2sbml working find

nguyenalter:
Hi @tritueviet,

We do support column comments in MySQL (implemented in https://github.com/holistics/dbml/pull/252)

Our parser currently does not recognize the `comment` word. We will fix this issue soon.

As a workaround, you can change the word `comment` to `COMMENT`.

tritueviet:
thanks u for support. It worked

PR #460, opened 2023-10-13
Fix comment syntax in MySQL parser

nguyenalter:
## Summary
Fix comment syntax in MySQL parser

## Issue
- https://github.com/holistics/dbml/issues/450
- #461 

- MySQL parser: Ignore case-sensitive when parsing `COMMENT` token
- Add a new testcase to test lowercase SQLs

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/src/parse/mysql/parser.pegjs
+++ b/packages/dbml-core/src/parse/mysql/parser.pegjs
@@ -456,8 +456,8 @@ ListOfNames = first:name rest:(comma _ name)*
 
 // Comment Syntax: Support "COMMENT 'this is a comment'"
 Comment 
-	= "COMMENT" _ ("=" _)? comment: "'" c:[^']* "'" {return c.join('')}
-	/ "COMMENT" _ ("=" _)? comment: "\"" c:[^"]* "\"" {return c.join('')}
+	= "COMMENT"i _ ("=" _)? comment: "'" c:[^']* "'" {return c.join('')}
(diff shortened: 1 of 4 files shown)


packages/dbml-core/src/parse/mysql/parser.pegjs
```diff
@@ -456,8 +456,8 @@ ListOfNames = first:name rest:(comma _ name)*
 
 // Comment Syntax: Support "COMMENT 'this is a comment'"
 Comment 
-	= "COMMENT" _ ("=" _)? comment: "'" c:[^']* "'" {return c.join('')}
-	/ "COMMENT" _ ("=" _)? comment: "\"" c:[^"]* "\"" {return c.join('')}
+	= "COMMENT"i _ ("=" _)? comment: "'" c:[^']* "'" {return c.join('')}
+	/ "COMMENT"i _ ("=" _)? comment: "\"" c:[^"]* "\"" {return c.join('')}
```
huyphung1602, line 460: Are there any issues if we ignore the case sensitive of this word? :thinking: Just want to learn it.

nguyenalter: There is no issue. We forgot to ignore the case sensitive for this token in the past :pray:

Most of the SQL keywords are case-insensitive so it's better if we can parse the keywords correctly.

huyphung1602: Try to create another Asana task to list the remaining keywords that tie in to these same problems. It could be brought up again for future updates.

huyphung1602 (approved):

Issue #470, opened 2023-10-29
Proposal: Make Ref syntax resilient to line breaks


AndreyTalanin:
### Objective:
Make it possible to break a `Ref` foreign key definition into multiple lines if needed.
### Syntax:
Currently it's only possible to define a foreign key in a single line:
```
Ref "FK_DependentTable_MainTable_ColumnName" {
    "schema"."DependentTable"."ColumnName" > "schema"."MainTable"."ColumnName"
}
```
The problem happens when there's a composite key with 3-4 columns:
```
Ref "FK_DependentTable_MainTable_MultipleColumns" {
    "schema"."DependentTable".("ColumnName1", "ColumnName2", "ColumnName3") > "schema"."MainTable".("ColumnName1", "ColumnName2", "ColumnName3")
}
```
With enough luck, it's possible to score for a 200-300 characters line, which is not comfortable to edit or even view.

The proposal is to allow breaking the line at least before or after the `<`/`>`/`<>` symbols:
```
Ref "FK_DependentTable_MainTable_MultipleColumns" {
    "schema"."DependentTable".("ColumnName1", "ColumnName2", "ColumnName3") >
        "schema"."MainTable".("ColumnName1", "ColumnName2", "ColumnName3")
}
```
Currently it is not possible and leads to a parsing error: 
> (Line:Character) Expected " " or valid name but "\n" found.

AndreyTalanin:
Well, current version of the DBML parser supports line breaks now. Closing this issue as not relevant anymore.

PR #458, opened 2023-10-12
Alternative dbml parser

huy-dna:
## Summary
* Implement an alternative dbml parser
* Doc: https://www.notion.so/holistics/DBML-Alternative-parser-6d3dff42a26d486eb70bdf519eac52a1?pvs=4
* Playground: https://dbml-explorer-demo-1.vercel.app/

## Issue
(issue link here)

## Lasting Changes (Technical)

(please list down: code changes/things that have wide-effect; new libraries/functions added that can be used by others; examples below)

* (Added `class EmailValidator` to validate email address' validity)
* (Added `Tenant#is_trial?` check)

## Checklist

Please check directly on the box once each of these are done

- [ ] Documentation (if necessary)
- [ ] Tests (integration test/unit test)
- [ ] Integration Tests Passed
- [ ] Code Review

PR #473, opened 2023-11-04
Replace 'sp+' with '__' around the relation sign in the ref body

AndreyTalanin:
## Summary
Replaced 'sp+' with '__' (requiring at least 1 whitespace) around the relation sign in the ref body, so it's possible to break a line before or after the relation sign for long `Ref` declarations representing composite foreign keys.

## Issue
#470

* Replaced 'sp+' with '__' around the relation sign in the ref body (line 181)
* Formatted the file on save, trimming trailing whitespaces (actually, unintentionally, can revert it and only leave the first change)

## Checklist

- [ ] Code Review

## P.S.
Please, let me know if I should continue with this PR and what else needs to be done (adding tests, etc.).

### DIFF
--- a/packages/dbml-core/src/parse/dbml/parser.pegjs
+++ b/packages/dbml-core/src/parse/dbml/parser.pegjs
@@ -40,7 +40,7 @@ expr
   }
   / __
 
-ProjectSyntax 
+ProjectSyntax
   = project name:(__ name)? _ "{" _ body:ProjectBody _ "}" {
     return {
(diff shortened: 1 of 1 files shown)


NQPhuc:
Hi @AndreyTalanin, 

Thank you for your interest in improving dbml.

However, we don't have to making change to the current dbml parse, as we're in the process of making a new parser for dbml (https://github.com/holistics/dbml/pull/458). Therefore, we'll consider your proposal at a later time.

Issue #212, opened 2021-09-22
Comments exported to MySQL using wrong quote type


djpeach:
The table definition:
```
Table Account {
  ID "int unsigned" [pk, increment, not null]
  
  Username varchar(255) [not null]
  Password varchar [not null]
  
  Note: 'The necessary details to identify and authenticate users'
}
```

The exported comment syntax: 
```ALTER TABLE `Account` COMMENT = "The necessary details to identify and authenticate users";```

Notice the double quotes on the comment string. MySQL expects single quotes.

**Comment by djpeach** [comment]
I also noticed the wrong quote type is used for default enums: 
```
`Status` ENUM ('Processing', 'Shipped', 'Delivered', 'Cancelled') NOT NULL DEFAULT "Processing";
```

nguyenalter:
Fixed in https://github.com/holistics/dbml/pull/477

PR #477, opened 2023-11-21
Fix: Export note string and default value

nguyenalter:
## Summary
- Fix exporter: Escape single quotes using double single quotes
  - Before: 
    - Note content: `A string contains '`
    - SQL note string: 
      - `A string contains "` (Postgres, SqlServer) 
      - `A string contains \'` (MySQL)
  - After:
    - Note content: `A string contains '`
    - SQL note string: `A string contains ''` (applied for all SQLs) 
   - Reason: Escaping a single quote in string literal using two single quotes is a standard way in SQL and it is supported in many modern DBMS (MySQL, Postgres, SqlServer, etc). See references below:
     - [Postgres](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS)
     - [MySQL](https://dev.mysql.com/doc/refman/8.2/en/string-literals.html#:~:text=A%20%27%20inside%20a%20string%20quoted%20with%20%27%20may%20be%20written%20as%20%27%27.)
     - [SQL Server (T-SQL)](https://learn.microsoft.com/en-us/sql/t-sql/data-types/constants-transact-sql?view=sql-server-ver16#character-string-constants:~:text=If%20a%20character%20string%20enclosed%20in%20single%20quotation%20marks%20contains%20an%20embedded%20quotation%20mark%2C%20represent%20the%20embedded%20single%20quotation%20mark%20with%20two%20single%20quotation%20marks)
 - Fix export DEFAULT value in MySQL: Use single quote instead of double quote
   - Before: `DEFAULT "value"`
   - After: `DEFAULT 'value'`
   - Reason: Following MySQL documentation: https://dev.mysql.com/doc/refman/8.2/en/data-type-defaults.html

## Issue
- https://github.com/holistics/dbml/issues/212

- Update MySQL, PostgreSQL and SqlServer exporters:
  - Escape single quote using two single quotes
  - MySQL: Use single quote instead of double quote when exporting default value of string type
- Update tests

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/__tests__/exporter/mssql_exporter/output/general_schema.out.sql
+++ b/packages/dbml-core/__tests__/exporter/mssql_exporter/output/general_schema.out.sql
@@ -74,7 +74,7 @@ GO
 
 EXEC sp_addextendedproperty
 @name = N'Table_Description',
-@value = 'This is a note in table "products"',
+@value = 'This is a note in table ''products''',
 @level0type = N'Schema', @level0name = 'dbo',
 @level1type = N'Table',  @level1name = 'products';
 GO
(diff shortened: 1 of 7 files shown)


huyphung1602:
Please help to update your description. It is long, but it is hard to understand what you are trying to explain. I especially do not understand the examples.

NQPhuc (approved):
LGTM!

Issue #481, opened 2023-11-28
v2 parser inconsistency with double-quoted list defaults


MattIPv4:
👋 Noticed this small inconsistency with the v2 parser vs. v1:

```
import { Parser, ModelExporter } from "@dbml/core";

const schema = `
Table foo {
  bar "varchar[]" [default: "{baz}"]
}
`;

const v1 = new Parser().parse(schema, 'dbml');
console.log(ModelExporter.export(v1.normalize(), 'postgres'));

const v2 = new Parser().parse(schema, 'dbmlv2');
console.log(ModelExporter.export(v2.normalize(), 'postgres'));
```

`v1` logs as expected, but `v2` throws `Can not find Enum field '{baz}'`.

https://replit.com/@MattIPv4/DBML-v2-parser-inconsistency

This is fixable by switching to a single-quote `'{baz}'` string, and looking at the docs, I think it should be using single-quotes as a plain-text string, but wanted to flag the parsing inconsistency all the same.

TeaNguyen:
Thank you for reporting the issue. We'll consider re-adding the double quotes shortly.

NQPhuc:
Fixed is https://github.com/holistics/dbml/pull/488

PR #488, opened 2024-01-04
Fix new DBML parser inconsistencies

NQPhuc:
### Resolve the following inconsistencies between old and new dbml parser behaviors:
1. Fix a bug causing multiple single quote to not report error.
```
Table user_role_in_diagram {
  role int [note: 'Role = sum(user's available permissions bit value)'] // should throw error on 's
}
```
2. Make keyworks check for default settings (true, false and null) case-insensitive.
3. Allow default value with double quote (https://github.com/holistics/dbml/issues/481)

4. Allow inline keywords in columns.
```
Table t {
  id int pk
}
```
5. Allow multiple column names on the same line in indexes definition.
```
Table citites {
  id integer pk
  name string

  indexes {
    id name 
  }
}
```
6. Allow array type in column.
```
Table users {
  id int
  order_ids int[]
}
```

* Improve error messages on validation errors
* Fix some errors reported with invalid error location (NaN on line and column)

## Issue
(issue link here)

* Ignore null attribute and setting bug cause by syntax error involve more than 2 single quotes
* Stop binding default indentifier to Enum
* Keywords check in binder and intepreter is now case-insensivity
* Update tests
* Refactor validator to not use the Template Method pattern
* Refactor Interpreter to make it modular

## Checklist

- [ ] Code Review


HuyDNA:
> 3. Unsupported: Array column types — due to the ambiguity in the grammar that this would introduce.
> 
> ```
> Table country {
>   id integer
>   cities string[] // not work
>   // equivalent: cities "string[]"
> }
> ```
This has been supported. However, there's a pitfall:

```
Table country[note: 'abc'] {
  ...
}
```
`country[note: 'abc']` would be treated as an array node, instead of separately as a name and a list, due to some ambiguity.
This can be avoided by adding a space between them.
It can indeed be solved completely but it would complicate the parser significantly. In my opinion, this error is unlikely to happen and we can yield more informative error messages for this pitfall so the user can be more aware? @NQPhuc

NQPhuc:
@HuyDNA I'm aware that since both table definition and column definition used the same rule (element-declaration), supporting array type for column would affect table definition as well.

I think for now we can accept the new inconsistency cause to table setting. **However, please help show a informative message to let the user know they need to leave empty space between table name and setting [].**

In the long run, we should consider breaking element-declaration into different declaration type depending, because as of right now, this is not very flexible.

HuyDNA:
@NQPhuc All consistencies have been resolved. Plus the following benefits:
* Error messages are more informative
* Better error markings: Example, all duplicate settings are marked, instead of only the last; All circular refs are marked, instead of only the last, etc. -> Better debugging
* The Interpreter is refactored into modules instead of being monolithic, adding other elements is easier.
* The Validator is more flexible, allowing catching more precise errors.

packages/dbml-parse/src/lib/analyzer/validator/utils.ts
```diff
@@ -25,47 +27,26 @@ import { createSchemaSymbolIndex } from '../symbol/symbolIndex';
...
-    case ElementKind.PROJECT:
+    case 'project':
       return ProjectValidator;
-    case ElementKind.REF:
+    case 'ref':
       return RefValidator;
-    case ElementKind.NOTE:
+    case 'note':
       return NoteValidator;
-    case ElementKind.INDEXES:
```
NQPhuc, line 68: @HuyDNA Please help re-add this constant nha. We shouldn't fixed the element type somewhere deep in the validator code.

packages/dbml-parse/src/lib/analyzer/validator/elementValidators/custom.ts
```diff
@@ -1,74 +1,75 @@
...
+  }
+
+  validate(): CompileError[] {
+    return [...this.validateContext(), ...this.validateName(this.declarationNode.name), ...this.validateAlias(this.declarationNode.alias), ...this.validateSettingList(this.declarationNode.attributeList), ...this.validateBody(this.declarationNode.body)];
+  }
 
-export default class CustomValidator extends ElementValidator {
-  protected elementKind: ElementKind = ElementKind.CUSTOM;
+  private validateContext(): CompileError[] {
+    if (this.declarationNode.parent instanceof ProgramNode || this.declarationNode.parent?.type?.value.toLowerCase() !== 'project') {
```
NQPhuc, line 25: Please use ElementKind here

packages/dbml-parse/src/lib/analyzer/validator/elementValidators/indexes.ts
```diff
@@ -1,88 +1,172 @@
...
+    this.publicSymbolTable = publicSymbolTable;
+    this.symbolFactory = symbolFactory;
+  }
+
+  validate(): CompileError[] {
+    return [...this.validateContext(), ...this.validateName(this.declarationNode.name), ...this.validateAlias(this.declarationNode.alias), ...this.validateSettingList(this.declarationNode.attributeList), ...this.validateBody(this.declarationNode.body)];
+  }
+
+  private validateContext(): CompileError[] {
+    if (this.declarationNode.parent instanceof ProgramNode || this.declarationNode.parent?.type?.value.toLowerCase() !== 'table') {
```
NQPhuc, line 38: Also here

packages/dbml-parse/src/lib/interpreter/elementInterpreter/project.ts
```diff
@@ -0,0 +1,78 @@
...
+          const errors = (new TableGroupInterpreter(sub, this.env)).interpret();
+          this.project.tableGroups!.push(this.env.tableGroups.get(sub)!);
+          return errors;
+        }
+        case 'enum': {
+          const errors = (new EnumInterpreter(sub, this.env)).interpret();
+          this.project.enums!.push(this.env.enums.get(sub)!);
+          return errors;
+        }
+        case 'note': {
```
NQPhuc, line 64: Should use elementKind here

huyphung1602 (approved):

Issue #492, opened 2024-01-17
[sql2dbml] Notes are not properly escaped


pierresouchay:
Consider this Porgresql SQL statement:

```
COMMENT ON TABLE public.my_bugs IS '
            For each row, a bug is described with reproductible input and user's input
        ';
```   
 
This comment would be translated into:
```
Note: 'For each row, a bug is described with reproductible input and user's input'
```
=> incorrect format, fails. We need to escape values within the notes first

PR #493, opened 2024-01-17
fix(sql2dbml): properly escape notes when notes do contain some '

pierresouchay:
## Summary
* Allow putting notes with simple quotes

## Issue
https://github.com/holistics/dbml/issues/492

Simply escape simple quotes in notes to avolid dbml parsers to fail

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/src/export/DbmlExporter.js
+++ b/packages/dbml-core/src/export/DbmlExporter.js
@@ -15,6 +15,10 @@ class DbmlExporter {
     return /\s*(\*|\+|-|\([A-Za-z0-9_]+\)|\(\))/g.test(str);
   }
 
+  static escapeNote (str) {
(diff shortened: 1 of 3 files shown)


packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -78,7 +78,7 @@ class DbmlExporter {
         constraints.push(value);
       }
       if (field.note) {
-        constraints.push(`note: '${field.note}'`);
+        constraints.push(`note: '${field.note.replaceAll("'", "\\'")}'`);
```
nguyenalter, line 81: You need to escape the table and enum notes. You can split this functionality to a util method.

pierresouchay: @nguyenalter done

PR #503, opened 2024-01-31
fix(notes): properly escape notes with new lines

pierresouchay:
escape new lines into notes

## Summary
* Comments in SQL can have carriage returns which break dbml since Notes are build with strings
* Escape those, so dbml can be used by other tools

## Issue
Relates to https://github.com/holistics/dbml/issues/492

### DIFF
--- a/packages/dbml-core/src/export/DbmlExporter.js
+++ b/packages/dbml-core/src/export/DbmlExporter.js
@@ -16,7 +16,7 @@ class DbmlExporter {
   }
 
   static escapeNote (str) {
-    return str.replaceAll("'", "\\'")
+    return str.replaceAll("'", "\\'").replaceAll("\n", "\\n");
   }
 
   static exportEnums (enumIds, model) {
--- a/packages/dbml-core/src/export/DbmlExporter.js
+++ b/packages/dbml-core/src/export/DbmlExporter.js
@@ -15,14 +15,14 @@ class DbmlExporter {
     return /\s*(\*|\+|-|\([A-Za-z0-9_]+\)|\(\))/g.test(str);
   }
 
-  static escapeNote (str, indentLengh) {
+  static escapeNote (str, indent) {
     if (str === null) {
       return ''
     }
     if (str.match(/[\n\r']/)) {
       const closure = function() {
         function compute_line(line, lineNumber){
-          var prefix = (lineNumber > 0) ? ' '.repeat(indentLengh) : "";
+          var prefix = (lineNumber > 0) ? ' '.repeat(indent + 3) : ""; // We add 3 because 3 simple quotes
           return prefix + line;
         }
         return compute_line
@@ -49,7 +49,7 @@ class DbmlExporter {
     return enumStrs.length ? enumStrs.join('\n') : '';
   }
 
-  static getFieldLines (tableId, model, indent) {
+  static getFieldLines (tableId, model) {
     const table = model.tables[tableId];
 
     const lines = table.fieldIds.map((fieldId) => {
@@ -98,16 +98,8 @@ class DbmlExporter {
         constraints.push(value);
       }
       if (field.note) {
-        var constraintPfxLen;
-        if (constraints.length == 0) {
-          // If we have no constraint yet, we have to add " [" before
-          constraintPfxLen = 2;
-        } else {
-          constraintPfxLen = ` [${constraints.join(', ')}]`.length + 1;
-        }
-        const notePfx = 'note: ';
-        const indentLengh = indent.length + line.length + constraintPfxLen + notePfx.length;
-        constraints.push(`${notePfx}${DbmlExporter.escapeNote(field.note, indentLengh)}`);
+        const indent = (line + ` [${constraints.join(', ')}]`).length + 7;
+        constraints.push(`note: ${DbmlExporter.escapeNote(field.note, indent)}`);
       }
 
       if (constraints.length > 0) {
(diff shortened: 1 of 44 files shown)


NQPhuc:
Hi @pierresouchay,

Thanks for contributing to dbml!

To help with getting this PR merged, kindly assist in:
* Include a test case (under packages/dbml-core/__tests__/exporter) for the problem you're trying to fix.
* In some systems, new lines is presented as `\n\r` so escaping \n if not enough (see attached video).
  * Besides this, I think we should export notes containing new lines into three double quotes `'''` syntax as it would be easier for users to review the imported note instead of having 1 long line of text separated by `\n`.

https://github.com/holistics/dbml/assets/11730168/f9c838f2-a41c-44ba-b810-add81317313f

pierresouchay:
> Hi @pierresouchay,
> 
> Thanks for contributing to dbml!
> 
> To help with getting this PR merged, kindly assist in:
> * Include a test case (under packages/dbml-core/__tests__/exporter) for the problem you're trying to fix.
> * In some systems, new lines is presented as `\r\n` so escaping `\n` is not enough (see attached video).
>   * Besides this, I think we should export notes containing new lines into three double quotes `'''` syntax as it would be easier for users to review the imported note instead of having 1 long line of text separated by `\n`.
> 
> https://github.com/holistics/dbml/assets/11730168/f9c838f2-a41c-44ba-b810-add81317313f
> 
> 

Yes. actually, there is a design issue IMHO, because the SQL parser does not properly escape SQL strings: currently, a comment or string with 2 simple quotes and/or with \n\r would not be read properly and keep the string as it is but remove \r and only output should add \n\r according to the current system. I tried to modify parsers in my previous MR, but since there are actually 2 parsers postgres legacy and the antl based), time for doing this was lacking and I failed.

I propose to keep it that way for now as changing the parser would require too much time (I also wanted to add proper support for functions), but currently, the existing parsers are a bit weak for real usage scenarios ;(

For instance, I had to implement a pre-processing phase to remove multilines functions.

I'll update the MR today to switch to triple simple quotes if ' or \n are detected in notes.

pierresouchay:
@NQPhuc all done!

pierresouchay:
@NQPhuc In order to have clean indentation, I had to revamp a bit a few things.

What I wanted is having this kind of annotation: https://github.com/holistics/dbml/pull/503/files#diff-b1ef04086803b098fb6b1b12c89129f332b8e6865b0803c7d82a620531696ad7R372

Something like:
```dbml
  "comments" "character varying(3850)" [note: '''Reviewer's comments
                                              WARNING: can have several lines!''']
```
=> with the next lines of comments being aligned with start of `'''`. I don't know if, in case of multiline, you prefer to have the `'''` being on the next line, tell me, and if you want, we can do the change to have instead this behaviour (with the last three simple comments on a new line):

```dbml
  "comments" "character varying(3850)" [note: '''Reviewer's comments
                                              WARNING: can have several lines!
                                              ''']
```



Since notes on lines, might be starting at different positions, I had to handle the indentation at varying places. I also took this opportunity to properly indent indices in tables, so `Indexes {` now are properly indented => which forced me to do lots of changes in UTests, that is why so many files have been modified (however, I did change only the ones needed to pass all unit tests, I did not change the dbml files that are inputs of UTests.

packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -15,8 +15,24 @@ class DbmlExporter {
...
+        function compute_line(line, lineNumber){
+          var prefix = (lineNumber > 0) ? ' '.repeat(indentLengh) : "";
+          return prefix + line;
+        }
+        return compute_line
+      }
+      const comp_line = closure();
+      const res = str.split(/[\r\n]/).map(comp_line).join("\n");
+      return `'''${res}'''`;
+    } else {
```
NQPhuc, line 33: I don't think we should prefix each new line with additional white-space because it would lead to weird formating, and will make markdown comment invalid as well.

packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -15,8 +15,24 @@ class DbmlExporter {
...
+      return `'${str}'`
+    }
   }
```
NQPhuc, line 36: * Since indentation is not yet used, I suggest to remove it to keep the code clean
* Should add back the escaping behavior, otherwise, we will, have to handle the edge case of ''' presented in the source text could break our note as well.
  * Beside that, in the old code, I found an edge case where if the input contain `\'`, our note will break
* In windows, \r\n (CRLF) is viewed as 1 line break, therefore, we should treat it as 1 line break only. 

```suggestion
  static escapeNote (str) {
    if (str === null) {
      return '';
    }

    // edge case: source text contain both \', we can preserve that by escaping any \ character
    let newStr = str.replaceAll('\\', '\\\\');
    newStr = newStr.replaceAll("'", "\\'");

    if (newStr.match(/[\n\r']/)) {
      newStr = newStr.replaceAll('\r\n', '\n'); // turn all CRLF to LF
      return `'''${newStr}'''`;
    }

    return `'${newStr}'`;
  }
```

NQPhuc (commented):
Thanks for taking the time to update your PR. There's only 1 remaining issue in the escapeNote() function, I have comment the suggestion for improvement. After that the PR should be good to go

pierresouchay:
@NQPhuc done

pierresouchay:
@NQPhuc I took your comment as it is... But it looks to me that according to spec, the single quote escaping is not working:
If there are some \n, it would escape ' while only ''' (triple single quotes) should be escaped according to https://dbml.dbdiagram.io/docs/#multi-line-string

So, I propose to switch to triple single quotes by default and only escape triple quotes within... What do you think?

pierresouchay:
@NQPhuc made the change as I find it better and avoid weird escape: we now switch to triple single quote for strings and only escape triple quotes as stated in specification. Also added some new UTests

NQPhuc (approved):
LGTM!

pierresouchay:
@NQPhuc Do you think you could merge it and release a new version?

Best Regards

PR #515, opened 2024-02-11
Fix validate when redundant args & `getMemberChain`

Huy-DNA:
## Summary
* Fix the problem:
  ```
    Enum Person {
       id integer
    }
  ```
  go uncaught in the validator
* The old bug was because the optional setting list was not treated properly.
* Fix the error message of invalid enum field.
* Fix the uncaught error in this situation:
   
image

  
## Issue
(issue link here)

* `getMemberChain` takes `ArrayNode` into account
* Enum and Ref validator

## Checklist

- [ ] Code Review


nguyenalter (approved):
LGTM 🚀

PR #526, opened 2024-02-21
Parser/fix `getRefId`

Huy-DNA:
## Summary
* Fix the `getRefId` function which causes some circular refs to slip
* The bug was because in the if clause, the comparison is string-based, and outside, the comparison is number-based.

## Issue
(issue link here)

## Checklist

- [ ] Code Review


NQPhuc (approved):
LGTM!

Issue #548, opened 2024-04-15
Postgresql importer does not take into account PK when set on ALTER TABLE statement


pierresouchay:
If we consider this Postgresql statement:

```sql
CREATE TABLE "countries" (
  "code" int,
  "name" varchar,
  "continent_name" varchar
);

ALTER TABLE ONLY "countries"
    ADD CONSTRAINT countries_pk PRIMARY KEY ("code");
```

=> The PK is not declared while it is for:

```sql
CREATE TABLE "countries" (
  "code" int PRIMARY KEY,
  "name" varchar,
  "continent_name" varchar
);
```

=> The PK is properly taken into account

=> alter on PK are not properly taken into accounts

I'll propose quickly a patch to provide such support

pierresouchay:
@NQPhuc the MR #549 is fixing the issue and adding lots of metadata to DBML output (as you can see in the test-cases)

PR #549, opened 2024-04-15
fix(postgresql): properly handle definition of unique and PK constraints in ALTER TABLE statements

pierresouchay:
Will fix https://github.com/holistics/dbml/issues/548

## Summary
* Implemented parsing of ALTER TABLE ... ADD CONSTRAINT for "UNIQUE|PRIMARY KEY"

## Issue
https://github.com/holistics/dbml/issues/548

This adds the constraints as indexes (with their names). If 1 single column is impacted, this will also add the attribute of the impacted column of the table model.

## Checklist

- [x] Code Review

## Summary by CodeRabbit

- **New Features**
	- Introduced a new table `comment_on_product` to enhance data structuring.
- **Enhancements**
	- Improved user data integrity by modifying the `email` field in the `users` table to be non-unique with a separate unique constraint.
	- Enhanced country identification by changing the `code` field in the `countries` table to have a dedicated primary key constraint.
- **Bug Fixes**
	- Fixed indexing issues by ensuring primary keys and unique constraints are correctly named and assigned.

### DIFF
--- a/packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
+++ b/packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
@@ -806,27 +806,29 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
             break;
           case TABLE_CONSTRAINT_KIND.UNIQUE:
           case TABLE_CONSTRAINT_KIND.INDEX:
-            if (cmd.value.columns == 0) break;
-            if (!(cmd.value.pk || cmd.value.unique)) break;
-            // eslint-disable-next-line prefer-destructuring
-            kind = cmd.kind;
-            const table = findTable(this.data.tables, schemaName, tableName);
-            if (!table) break;
-            // multi column constraint -> need to create new index
-            if (cmd.value.columns.lenght > 1) {
-              const index = new Index({
-                name: cmd.value.name,
-                columns: cmd.value.columns,
-              });
-              index[cmd.value.unique ? 'unique' : 'pk'] = true;
-              table.indexes.push(index);
+            {
+              if (cmd.value.columns.length === 0) break;
+              if (!(cmd.value.pk || cmd.value.unique)) break;
+              // eslint-disable-next-line prefer-destructuring
+              kind = cmd.kind;
+              const table = findTable(this.data.tables, schemaName, tableName);
+              if (!table) break;
+              // multi column constraint -> need to create new index
+              if (cmd.value.columns.lenght > 1) {
+                const index = new Index({
+                  name: cmd.value.name,
+                  columns: cmd.value.columns,
+                });
+                index[cmd.value.unique ? 'unique' : 'pk'] = true;
+                table.indexes.push(index);
+                break;
+              }
+              // single column contraint -> update the column instead
+              const key = cmd.value.columns[0].value;
+              const fieldToUpdate = table.fields.find(f => f.name === key);
+              if (fieldToUpdate) fieldToUpdate[cmd.value.unique ? 'unique' : 'pk'] = true;
               break;
             }
-            // single column contraint -> update the column instead
-            const key = cmd.value.columns[0].value;
-            const fieldToUpdate = table.fields.find(f => f.name === key);
-            if (fieldToUpdate) fieldToUpdate[cmd.value.unique ? 'unique' : 'pk'] = true;
-            break;
           default:
             break;
         }
--- a/packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
+++ b/packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
@@ -810,6 +810,9 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
             if (cmd.value.pk || cmd.value.unique) {
               let set_prop = cmd.value.unique ? (obj) => obj["unique"] = true : (obj) => obj["pk"] = true;
               const table = findTable(this.data.tables, schemaName, tableName);
+              if (!table){
+                break;
+              }
               if (cmd.value.columns.length == 1) {
                 // We are in the single column case, we can update the column as well
                 let key = cmd.value.columns[0].value;
--- a/packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
+++ b/packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
@@ -806,33 +806,26 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
             break;
           case TABLE_CONSTRAINT_KIND.UNIQUE:
           case TABLE_CONSTRAINT_KIND.INDEX:
+            if (cmd.value.columns == 0) break;
+            if (!(cmd.value.pk || cmd.value.unique)) break;
+            // eslint-disable-next-line prefer-destructuring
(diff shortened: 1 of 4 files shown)


packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
```diff
@@ -804,6 +804,37 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
...
+                index = new Index({
+                  name: cmd.value.name,
+                  columns: cmd.value.columns,
+                });
+                table.indexes.push(index);
+              }
+              set_prop(index);
+              index.columns = cmd.value.columns;
+            }
+            break;
```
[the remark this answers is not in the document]
pierresouchay: done

pierresouchay:
@NQPhuc I checked the changes in line by line in [packages/dbml-core/__tests__/importer/postgres_importer/output/db_dump.out.dbml](https://github.com/holistics/dbml/pull/549/files#diff-b1ef04086803b098fb6b1b12c89129f332b8e6865b0803c7d82a620531696ad7) => and everything looks fine (it adds lots of pk that were missing from initial test-case)

pierresouchay:
@NQPhuc I also had to fix to let pk have names in PyDBML: https://github.com/Vanderhoof/PyDBML/pull/35

=> Any comment on that? is
```dbml
Table{
  column1 int
  Index {
    column1 [pk, name: 'column1_idx']
  }
}
```
Is it a correct statement according to spec? For me, yes.

packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
```diff
@@ -804,6 +804,36 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
...
```
NQPhuc, line 836: Hi I suggest to update the logic to the below, the main points are:
* Generating both fields settings and indexes for the same constraint is not recommended, because when re-exported to SQL, it'll generate both a column constraint in CREATE TABLE column definition and a CREATE INDEX statement. Therefore. we should only choose one.
* IMO, using pk/unique field settings to describe SQL constraints is the better because it more inline with concept of SQL constraints, while indexes fit more with CREATE INDEX.. statements. However, it has two downsides:
  * It can only used for single column constraint -> I agree that for multi-column constraint, using to indexes is our only choice.
  * We can't name the pk/unique constraint in field settings -> For now, I recommend that for single-column named constraint, we still import it as a field property. In the future, we intend to add the ability to name field pk/unique property through this syntax in dbml. Please let me know what you think about this:
  ```
  table a {
    id int [pk: "pk_constraint_name"]
  }
  ...
  ```
```suggestion
          case TABLE_CONSTRAINT_KIND.UNIQUE:
          case TABLE_CONSTRAINT_KIND.INDEX: {
            if (!(cmd.value.pk || cmd.value.unique)) break;

            // eslint-disable-next-line prefer-destructuring
            kind = cmd.kind;

            const table = findTable(this.data.tables, schemaName, tableName);

            if (!table) break;

            // multi column constraint -> need to create new index
            if (cmd.value.columns.lenght > 1) {
              const index = new Index({
                name: cmd.value.name,
                columns: cmd.value.columns,
              });

              index[cmd.value.unique ? 'unique' : 'pk'] = true;

              table.indexes.push(index);

              break;
            }

            // single column contraint -> update the column instead
            const key = cmd.value.columns[0].value;
            const fieldToUpdate = table.fields.find(f => f.name === key);
            if (fieldToUpdate) fieldToUpdate[cmd.value.unique ? 'unique' : 'pk'] = true;

            break;
          }
```

NQPhuc:
> @NQPhuc I also had to fix to let pk have names in PyDBML: [Vanderhoof/PyDBML#35](https://github.com/Vanderhoof/PyDBML/pull/35)
> 
> => Any comment on that? Does it sound right to you?
> 
> ```
> Table my_table{
>   column1 integer
>   Index {
>     column1 [pk, name: 'column1_idx']
>   }
> }
> ```
> 
> Is it a correct statement according to spec? For me, yes.

@pierresouchay  Yes, for named pk indexes, this is correct according to spec. (minor thing but it should be `indexes` instead of `index`. But I believe that's just a typo)

packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
pierresouchay: @NQPhuc Ok, I am gonna take your path, I am gonna update this today (and update the test cases)

1. if constraint is on a single column, use "pk"/unique in column (without any name)
2. If constraint is on multiple columns, use "Indexes"

pierresouchay:
> @pierresouchay Yes, for named pk indexes, this is correct according to spec. (minor thing but it should be `indexes` instead of `index`. But I believe that's just a typo)

Indeed, was a typo: Index should be Indexes

packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
pierresouchay: @NQPhuc  OK, done your way, this is ready!

NQPhuc: Thanks, I think this PR is almost done. There's only one thing left: can you help add some test for ALTER TABLE ADD CONSTRAINT with multi-column to make sure it working properly?

packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
```diff
@@ -804,6 +804,29 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
             cmd.value.endpoints[0].schemaName = schemaName;
             this.data.refs.push(cmd.value);
             break;
+          case TABLE_CONSTRAINT_KIND.UNIQUE:
+          case TABLE_CONSTRAINT_KIND.INDEX:
+            if (cmd.value.columns == 0) break;
```
NQPhuc, line 809: I think you mean columns.length? 🤔 
```suggestion
            if (cmd.value.columns.length === 0) break;
```

packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
```diff
@@ -804,6 +804,29 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
...
+            if (!(cmd.value.pk || cmd.value.unique)) break;
+            // eslint-disable-next-line prefer-destructuring
+            kind = cmd.kind;
+            const table = findTable(this.data.tables, schemaName, tableName);
```
NQPhuc, line 813: Declaring these without scoping them will give a lot of lint error, and potentially cause conflict with same name `const` in other cases. Please help enclosed the this whole code from  line 809 to line 829 within brackets { }. Thanks

pierresouchay:
@NQPhuc we introduced a regression in changes when not doing systematically indexes. Fixed that (typo).

Also added new tests to take into account multiple indexes. 

Also changed scoping of variables.

Should be good now

packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
pierresouchay: fixed

PR #579, opened 2024-06-11
Fix/silent errors in compiler api stack container

Huy-DNA:
## Summary
* Fix silent error in this case:
image

* The reason is that:
1️⃣  `Compiler.container.token` return the offset in the flat token stream (the stream with all tokens and invalid, trivial tokens in the same level in a list.
2️⃣  `Compiler.container.stack` use the nested token stream (the stream with all nontrivial and valid tokens only, the invalid and trivial tokens are nested inside of these) -> Mismatch index and list -> index out of bound.

## Issue
(issue link here)

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-parse/src/compiler.ts
+++ b/packages/dbml-parse/src/compiler.ts
@@ -192,7 +192,7 @@ export default class Compiler {
       (offset: number): readonly Readonly<SyntaxNode>[] => {
         const tokens = this.token.flatStream();
         const { index: startIndex, token } = this.container.token(offset);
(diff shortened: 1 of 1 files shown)


packages/dbml-parse/src/compiler.ts
huyphung1602, line 1: IMO, we should not import the whole lodash lib em. Please help to update the code to import the methods you want to use. Example,
```javascript
import { map } from 'lodash';
```

huyphung1602 (approved):
LGTM, please help to re-check this update with dbdiagram nha @nguyenalter

PR #588, opened 2024-07-16
fix: safe unwrapping in suggesting attribute value when name is blank

Huy-DNA:
## Summary
* Fix this error:

https://github.com/user-attachments/assets/2487ff80-461d-445f-aabf-65d6e38963e1

* This is due to when the attribute name is blank and the cursor is currently after the colon, the autocompletion provider will try to unwrap the non existent attribute name (which is `None`) to get the suggestions for the attribute value.

## Issue
(issue link here)

* `suggestInAttribute` in completion provider: `unwrap_or` instead of `unwrap`

## Checklist

- [ ] Code Review


huyleminh01 (approved):
LGTM!

PR #592, opened 2024-07-17
chore: config vite to gen .d.ts file on build

Huy-DNA:
## Summary
* Config vite to gen .d.ts file on build
* Aid type information when using this package

## Issue
(issue link here)

## Checklist

- [ ] Code Review


huyleminh01:
I found this error when trying to run the code. Please fix it first.
image

packages/dbml-parse/package.json
```diff
@@ -22,7 +23,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "vite build",
+    "build": "tsc --noEmit && vite build",
```
huyleminh01, line 26: You added a package to generate the definition file, and does it have type checking?

If yes, why do you have to add `tsc --noEmit` here?

Huy-DNA: It's a common idiom anh :v In this section here https://vitejs.dev/guide/features#transpile-only 

image

Huy-DNA: By the way, the package `vite-dts` is a build-only package to gen `.d.ts` file :V this would help providing autocompletion when developing other packages that rely on `@dbml/parse` (the language service would lookup the `.d.ts` files to provide type checking).

NQPhuc: So because vite only emit transpile the code to `d.ts files`, it doesn't perform type-checking during `vite build`. That's why you add `tsc --noEmit` to type-check beforehand?

NQPhuc (approved):

packages/dbml-parse/package.json
Huy-DNA: ye anh, vite only builds anh, so tsc --noEmit is used to perform typecheck before build.

PR #594, opened 2024-07-23
Parse/fix!/no longer support escaping triple quote

Huy-DNA:
## Summary
* BREAKING CHANGE: No longer support escaping triple quote

## Issue
(issue link here)

## Checklist

- [ ] Code Review


packages/dbml-parse/tests/lexer/input/unclosed_strings.in.dbml
```diff
@@ -4,6 +4,4 @@
 
 '''This is also an unclosed string
 
-\''''''
```
NQPhuc, line 7: Please help add at least one more test for case of `'''... \''''` to show that it work

NQPhuc (commented):
To account for this change please also help:
* Update the dbml syntax docs
* Update the dbml exporter to:
  * Remove escape of triple quotes
  * Always escape single quotes

dbml-homepage/docs/docs.md
```diff
@@ -454,7 +454,8 @@ Note: '''
 - Line continuation: `\` backslash
 - Escaping characters:
   - `\`: using double backslash `\\`
-  - `'''`: using `\'''`
+  - `'`: using `\'`
+  - `"`: using `\"`
```
NQPhuc, line 458: I don't remember we ever have this rule, it this behavior newly introduced in this PR or we just didn't document them before?

Huy-DNA: we didn't document this before anh, I think because we explicitly escape `'`, we should mention this rule.

NQPhuc (approved):
LGTM!

PR #596, opened 2024-07-25
Add database connectors

huyphung1602:
## Summary
Add database connectors:
- Postgres
- MSSQL
- MySQL

Support fetching database schema JSON and converting it to DBML.

## Issue
(issue link here)

* Add database connectors: postgres, mssql, and mysql (wip)
* Add `databaseGenerator` to generate the database object from schema JSON.
* Add new CLI command: `db2dbml`

## Checklist

- [ ] Code Review


packages/dbml-cli/src/cli/index.js
```diff
@@ -39,7 +40,27 @@ function sql2dbml (args) {
...
+  program
+    .usage('[options]')
+    .option('--postgres', 'connect to a postgresql database')
+    .option('--database <name>', 'database name')
+    .option('--host <host>', 'database host. default: localhost')
+    .option('--port <port>', 'database port. default: 5432')
+    .option('--user <user>', 'database user')
+    .option('--password <password>', 'database password')
+    .option('--connection-string <connectionString>', 'database connection string')
+    .option('-o, --out-file <pathspec>', 'compile all input files into a single files');
```
NQPhuc, line 55: We can remove the need to use `--connection-string` before the connection string by doing this:

```
program
    .usage('<connection-string> [options]')
    .arguments('database connection string')
    .option(...)
    ....
```

And then in `connectionHandler`, you can read the connection-string through the programs.args (it is an array)

.github/workflows/test.yml
```diff
@@ -12,13 +12,32 @@ on:
 jobs:
   build:
     runs-on: ubuntu-22.04
-
     strategy:
       matrix:
         node-version: [16.x]
+    services:
+      postgres:
+        image: postgres:11.21-bullseye
```
NQPhuc, line 20: Is there any reason we perform this test on postgres 11 and not the newest postgres 15? If our goal is to make sure db2dbml work on all version from 11->15, I think we should test both

packages/dbml-core/src/parse/connection/postgres/utils.js
```diff
@@ -0,0 +1,519 @@
...
+  Endpoint,
+  Enum,
+  Field,
+  Index,
+  Table,
+  Ref,
+} from '../../ANTLR/ASTGeneration/AST';
+
+const connectPg = async (connection) => {
+  const client = new Client(connection);
```
NQPhuc, line 13: When trying to connect without providing a password, I encounter this error, and it hang the process for me.

Screenshot from 2024-07-31 11-45-58

I think this is an uncatchable error in pg and you should provide the connection arguments a default password to avoid this error.

packages/dbml-core/src/connector/index.js
```diff
@@ -0,0 +1,13 @@
+import DatabaseConnector from '../parse/DatabaseConnector';
+import ModelExporter from '../export/ModelExporter';
+
+async function _fetch (connection, format) {
```
NQPhuc, line 4: Are you adding _fetch to avoid collision with the global [fetch function in node 18](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch)? If yes then I don't understand why you are re-exporting it as `fetch`. Have you consider a different name?

packages/dbml-core/src/parse/Parser.js
```diff
@@ -54,6 +54,10 @@ class Parser {
     return new Parser().parse(str, format);
   }
 
+  static async fetchPostgresToJSON (connection) {
+    return fetch(connection, 'postgres');
```
NQPhuc, line 58: As mention above, please help test and make sure this work without problem on node 18, otherwise, we'll have to use a different name

packages/dbml-core/src/parse/connection/postgres/utils.js
huyphung1602: We switched to use connection string => we do not care about this issue again.

packages/dbml-core/src/parse/Parser.js
NQPhuc: @huyphung1602 Remember to change this to `fetchSchema`

NQPhuc (approved):

PR #602, opened 2024-08-06
Add MySQL connector

huyleminh01:
## Summary

- Add connector for fetching information schema from MySQL.

## Issue
(issue link here)

## Checklist

- [ ] Code Review


packages/dbml-core/src/connectors/mysqlConnector.js
```diff
@@ -0,0 +1,509 @@
...
+  if (extraType === 'VIRTUAL GENERATED') {
+    return `GENERATED ALWAYS AS (${generationExpression}) VIRTUAL`;
+  }
+
+  if (extraType === 'STORED GENERATED') {
+    return `GENERATED ALWAYS AS (${generationExpression}) STORED`;
+  }
+
+  // for timestamp data type
+  if (extraType.includes('on update CURRENT_TIMESTAMP')) {
```
huyphung1602, line 40: I think checking the CURRENT_TIMESTAMP is not enough. For example: NOW(), INTERVAL, ...

packages/dbml-core/src/parse/databaseGenerator.js
```diff
@@ -74,7 +74,7 @@ const createIndexes = (rawIndexes) => {
 const createTables = (rawTables, rawFields, rawIndexes, tableConstraints) => {
   return rawTables.map((rawTable) => {
     const { name, schemaName, note } = rawTable;
-    const key = `${schemaName}.${name}`;
+    const key = `${schemaName || ''}.${name}`;
```
huyphung1602, line 77: I think it should be changed to:
```javascript
const key = schemaName ? `${name}` : `${schemaName}.${name}`;
```

packages/dbml-core/src/connectors/mysqlConnector.js
huyleminh01: Ye, I'll add more test cases for this a.
FYI, the checking value is defined by mysql like this

packages/dbml-core/src/parse/databaseGenerator.js
huyleminh01: Fixed in https://github.com/holistics/dbml/pull/602/commits/f0dc9d52059958f8e1042c98657b695762d8ef2c

packages/dbml-core/src/connectors/mysqlConnector.js
huyleminh01: I added more test in this commit https://github.com/holistics/dbml/pull/602/commits/f0dc9d52059958f8e1042c98657b695762d8ef2c

Also, the `on update` clause can be used for datetime or timestamp; possible values are `CURRENT_TIMESTAMP` and its alias (`now()`, `localtime()`, etc.)

huyphung1602: So, you should also check the auto-increment attribute?

huyleminh01: Actually, I checked it in this section 
image
Since it is usefull for the increment attribute of DBML field.

huyphung1602 (approved):

PR #603, opened 2024-08-07
Add mssql connector

huyphung1602:
## Summary
* Add MSSQL connector to convert from MSSQL schema to DBML code.

## Issue
(issue link here)

* Add code to fetch and convert MSSQL schema to DBML code
* Fix quote when dbdefault is string type. (using ' or ''')
* Add tests and fixed related tests
* Handle the composite FKs

## Checklist

- [ ] Code Review


packages/dbml-cli/src/connectors/MssqlConnector.js
```diff
@@ -0,0 +1,508 @@
...
+
+    if (!fields[table_name]) fields[table_name] = [];
+    const field = generateField(row);
+    if (enumValues) {
+      field.type = {
+        type_name: check_constraint_name,
+        schemaName: table_schema,
+      };
+    }
+    fields[table_name].push(field);
```
NQPhuc, line 239: Will this work the case where we have the many table with the same name but in multiple schema? From reading the code, I believe we can't distinguish which `schema.table` the field belong to later on when we passed the `fields array` to `DatabaseGenerator`. For this to work, I think you could use a key like: `schemaName.tableName`

Also please check for this case with your postgres connector as well.

packages/dbml-cli/__test__/cli.test.js
```diff
@@ -34,6 +34,7 @@ describe('@dbml/cli', () => {
       const fileNames = fs.readdirSync(path.join(dirName, './out-files'));
       let content = fs.readFileSync(path.join(dirName, './out-files', fileNames[0]), 'utf-8');
       content = content.replace(/--.*(?:\n)*/g, '');
+      content = content.replaceAll('\r\n', '\n');
```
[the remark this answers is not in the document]
huyphung1602: Still need nha. The below regex is used to remove the `-- comments` out of the generated sql file. I added the condition to make sure this LOC only run when testing dbml2sql command

packages/dbml-cli/src/connectors/MssqlConnector.js
huyphung1602: Fix this issue for both postgresql and mssql nha.

NQPhuc (approved):

PR #623, opened 2024-08-26
Add Snowflake connector

huyphung1602:
## Summary
* Add Snowflake connector

## Issue
(issue link here)

* Add Snowflake connector
* Allow users to select schema when fetching schema information from: Postgresql, Mysql, MSSQL, and Snowflake.

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-connector/src/connectors/snowflakeConnector.ts
+++ b/packages/dbml-connector/src/connectors/snowflakeConnector.ts
@@ -335,7 +335,6 @@ const fetchSchemaJson = async (connection: string): Promise<DatabaseSchema> => {
   }
   // Schemas: schema1,schema2,schema3
   const schemas = config.SCHEMAS ? parseSchema(config.SCHEMAS) : [];
-  console.log(schemas);
   const databaseName = config.DATABASE;
 
   const tablesAndFieldsRes = generateTablesAndFields(conn, schemas);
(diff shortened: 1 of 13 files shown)


packages/dbml-connector/src/connectors/snowflakeConnector.ts
```diff
@@ -0,0 +1,370 @@
...
+  DefaultType,
+  Field,
+  FieldsDictionary,
+  Index,
+  IndexesDictionary,
+  Table,
+  TableConstraintsDictionary,
+} from './types';
+
+type constraintRow = {
```
huyleminh01, line 24: Types and Interfaces should be in PascalCase a

packages/dbml-connector/src/connectors/snowflakeConnector.ts
```diff
@@ -0,0 +1,370 @@
...
+  return {
+    indexes,
+    tableConstraints,
+  };
+};
+
+const fetchSchemaJson = async (connection: string): Promise<DatabaseSchema> => {
+  const config = parseConnectionString(connection);
+  const conn = await connectToSnowflake(config);
+  if (conn instanceof Error) {
```
huyleminh01, line 332: Does snowflake returns error instead of throwing it?
If it throws the error, you do not need to check here

huyleminh01 (approved):
LGTM

PR #629, opened 2024-09-05
Fix: When the parameter format is JSON in Parser.parse, the first parameter 'parse' must be RawDatabase

Mrxyy:
## Summary
The Database.constructor parameter requires RawDatabase because the Database.export method executed by ModelExporter.export (format=json) returns a regular object instead of RawDatabase, which is different from RawDatabase. Therefore, when using Parser.parse(str, format=json), str must be in RawDatabase or its json form to alert everyone about unexpected errors caused by their differences.

image

The Database.constructor parameter requires RawDatabase 
imageimage

## Issue
(issue link here)

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/types/parse/Parser.d.ts
+++ b/packages/dbml-core/types/parse/Parser.d.ts
@@ -13,6 +13,6 @@ declare class Parser {
      * Should use parse() instance method instead of this static method whenever possible
      */
     static parse(str: string, format: 'mysql' | 'postgres' | 'dbml' | 'dbmlv2' | 'schemarb' | 'mssql' | 'json'): Database;
-    parse(str: string, format: 'mysql' | 'postgres' | 'dbml' | 'dbmlv2' | 'schemarb' | 'mssql' | 'json'): Database;
+    parse(str: string | RawDatabase, format: 'mysql' | 'postgres' | 'dbml' | 'dbmlv2' | 'schemarb' | 'mssql' | 'json'): Database;
 }
 export default Parser;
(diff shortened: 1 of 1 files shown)


packages/dbml-core/types/parse/Parser.d.ts
```diff
@@ -13,6 +13,6 @@ declare class Parser {
      * Should use parse() instance method instead of this static method whenever possible
      */
     static parse(str: string, format: 'mysql' | 'postgres' | 'dbml' | 'dbmlv2' | 'schemarb' | 'mssql' | 'json'): Database;
-    parse(str: string, format: 'mysql' | 'postgres' | 'dbml' | 'dbmlv2' | 'schemarb' | 'mssql' | 'json'): Database;
+    parse(str: string | RawDatabase, format: 'mysql' | 'postgres' | 'dbml' | 'dbmlv2' | 'schemarb' | 'mssql' | 'json'): Database;
```
Huy-DNA, line 16: Hi @Mrxyy, thanks for your interest in contributing to DBML :bow:, about this, can you use [overload function declaration](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-function-d-ts.html) in Typescript to make the signature more precise? We only allow `str` to be `RawDatabase` when `format` is `json`.

Mrxyy: Thank you for your suggestion.

Huy-DNA:
@NQPhuc Looks good to me, pls help check again and approve nhe a :eyes:

NQPhuc (approved):
LGTM!

Issue #631, opened 2024-09-06
Missing types.ts files in @dbml/parse@3.8.0


Alfoussen:
Hi,

Some files are absent from @dbml/parse@3.8.0, failing the typescript compilation

```typescript
import { Parser } from '@dbml/core';
```

```shell
$ npx tsc -b 

node_modules/@dbml/parse/dist/lib/lexer/tokens.d.ts:1:26 - error TS2307: Cannot find module '../types' or its corresponding type declarations.

1 import { Position } from '../types';
                           ~~~~~~~~~~

node_modules/@dbml/parse/dist/lib/parser/nodes.d.ts:3:26 - error TS2307: Cannot find module '../types' or its corresponding type declarations.

3 import { Position } from '../types';
                           ~~~~~~~~~~

node_modules/@dbml/parse/dist/compiler.d.ts:8:26 - error TS2307: Cannot find module './lib/interpreter/types' or its corresponding type declarations.

8 import { Database } from './lib/interpreter/types';
```

<img width="590" alt="image" src="https://github.com/user-attachments/assets/d65220b1-66fc-44b4-b1b7-c051010a1be6">

huyleminh01:
Hi @Alfoussen ,

Could you provide use with more information about your current situation?

- Typescript version
- Your `tsconfig.json` (if possible)

Thank you.

Alfoussen:
@huyleminh01 

Typescript : Version 5.5.4

tsconfig.json
```json
{
    "compilerOptions": {
      "module": "commonjs",
      "esModuleInterop": true,
      "target": "es6",
      "moduleResolution": "node",
      "sourceMap": true,
      "outDir": "dist"
    },
    "include": ["src/cli/**/*"],
    "exclude": ["**/*.spec.ts"],
    "lib": ["es2015"]
  }
```

Alfoussen:
Hi @huyleminh01 

I manually added those missing files named type.d.ts and could make it compile. I just wonder why they are absent (or suppressed) from the original package... I escalated back to some older versions and still same missing files. Let me know if you need more info.

Huy-DNA:
hi @Alfoussen, thanks for letting us know this :heart: We've acknowledged the problem and will notify you when the problem is fixed!

Huy-DNA:
@Alfoussen  fixed in https://github.com/holistics/dbml/pull/632, please let us know if you encounter any other errors.

PR #632, opened 2024-09-07
chore/upgrade vite plugin dts & fix tsconfig to properly include all sources

Huy-DNA:
## Summary
* Upgrade `vite-plugin-dts`, which resolves this [security issue](https://github.com/holistics/dbml/security/dependabot/413).
   
image

* Fix `tsconfig` to `include` all `src` not just the entry file, resolving [this issue](https://github.com/holistics/dbml/issues/631).

image

* Remove extraneous `vite-dts` (intended `vite-plugin-dts`)

## Issue
(issue link here)

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-parse/vite.config.ts
+++ b/packages/dbml-parse/vite.config.ts
@@ -6,9 +6,7 @@ import dts from 'vite-plugin-dts';
 
 export default defineConfig({
   plugins: [
-    dts({
-      insertTypesEntry: true,
-    }),
+    dts(),
   ],
   build: {
     lib: {
--- a/packages/dbml-parse/package.json
+++ b/packages/dbml-parse/package.json
@@ -38,7 +38,7 @@
     "typescript": "^5.2.2",
     "vite": "^4.5.0",
     "vite-dts": "^1.0.4",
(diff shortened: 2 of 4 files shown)


packages/dbml-parse/vite.config.ts
```diff
@@ -6,9 +6,7 @@ import dts from 'vite-plugin-dts';
 
 export default defineConfig({
   plugins: [
-    dts({
-      insertTypesEntry: true,
-    }),
+    dts(),
```
nguyenalter, line 9: Do we still need the [types option](https://github.com/holistics/dbml/blob/4a4f1a954f551b345fc5f33bdd91af9826a99712/packages/dbml-parse/package.json#L15) in package.json file after removing the `insertTypesEntry: true` option?

Huy-DNA: Yeah anh :v that `insertTypeEntry` is redundant for `dts` anh, I remove that and it still generates `index.d.ts`, so `types` in `package.json` still points to that

nguyenalter (approved):

PR #634, opened 2024-09-10
Fix postgres connector issues

huyphung1602:
## Summary
Fix postgres connector issues related to:
- Duplicated fields
- Failed to fetch enums if there is the same enum_type name in different schemas.
Fix wrong enums when fetching in mssql.

## Issue
(issue link here)

- Fix the eslintrc to parse ts file correctly
- Remove wrong enums when fetching in mssql
- Update the sql which fetching tables, fields, and comments to  avoid fetching duplicated fields
- Update the sql which fetching enums to support the same enum_type name in different schema_name

## Checklist

- [ ] Code Review

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,44 +1,56 @@
 {
+  "root": true,
   "env": {
-      "es6": true,
-      "node": true,
-      "jest/globals": true
+    "es6": true,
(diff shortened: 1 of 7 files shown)


huyleminh01 (approved):
LGTM

Issue #644, opened 2024-10-08
Field size in dbml file


akuiteo-dbdocs:
Before using the automatic generation of the DBML file for SQL server, I had queries allowing me to generate my file in DBML format.
In these scripts, I retrieved the field type and size, which was very useful when viewing fields in dbdocs. 

Example : 
```
Table "dbo"."B_BATCH_PLANIF" {
  "BBPL_ANNEE" numeric(14,2)
  "BBPL_AV_ANNEE" nvarchar(60) [note: 'Mode CRON : Annee(s)']
  "BBPL_AV_HEURE" nvarchar(10) [note: 'Mode CRON : Heures']
}
```

With “dbdocs db2dbml mssql” I no longer have this information, which is a big loss compared with what I used to produce. Would it be possible to have this data? 

```
Table "dbo"."B_BATCH_PLANIF" {
  "BBPL_ANNEE" numeric
  "BBPL_AV_ANNEE" nvarchar [note: 'Mode CRON : Annee(s)']
  "BBPL_AV_HEURE" nvarchar [note: 'Mode CRON : Heures']
}
```

huyleminh01:
Hi @akuiteo-dbdocs,

Thank you for reporting the bug to us.

It maybe a bug related to db2dbml command, we will investigate this issue and get back to you later.

Thanks.

huyleminh01:
Hi @akuiteo-dbdocs ,

We checked and confirmed that our connector does not fetch the field size _(numeric scale, numeric precision, and string length)_.

So we have released a fixed version in `@dbml/connector@3.9.1`. We will update it to dbdocs cli later. But you can try [dbml cli ](https://dbml.dbdiagram.io/cli#generate-dbml-directly-from-a-database)to test our new fix first.

Thanks.

huyleminh01:
Hi @akuiteo-dbdocs ,

There is good news: we've just released [`dbdocs cli v0.13.1`](https://www.npmjs.com/package/dbdocs) contains this fix 🥳 . Please try it out

akuiteo-dbdocs:
It's work like a charm ! Thanks a lot

PR #645, opened 2024-10-10
Fix mssql connector - Missing data type size for numeric-based and string-based columns

huyleminh01:
## Summary
- Fetch numeric precision and scale for numeric-based column
- Fetch character max length for string-based column

## Issue
https://github.com/holistics/dbml/issues/644

## Checklist

- [x] Code Review

### DIFF
--- a/packages/dbml-cli/__test__/db2dbml/mssql/expect-out-files/schema.dbml
+++ b/packages/dbml-cli/__test__/db2dbml/mssql/expect-out-files/schema.dbml
@@ -55,6 +55,19 @@ Table "dbo"."NumberTypes" {
   "TINYINTCol" tinyint [default: 0]
 }
 
+Table "dbo"."NumberTypesNoDefault" {
+  "BIGINTCol" bigint
+  "BITCol" bit
+  "DECIMALCol" decimal(10,2)
+  "FLOATCol" float
+  "ID" int [pk, not null, increment]
+  "INTCol" int
+  "NUMERICCol" numeric(10,2)
+  "REALCol" real
+  "SMALLINTCol" smallint
+  "TINYINTCol" tinyint
+}
+
 Table "dbo"."ObjectTypes" {
   "BinaryField" binary [default: `0x00`]
   "Id" int [pk, not null, increment]
@@ -91,7 +104,7 @@ Table "dbo"."order_items" {
   "order_item_id" int [pk, not null, increment]
   "product_id" int [unique, not null]
   "quantity" int [not null]
-  "unit_price" decimal [not null]
+  "unit_price" decimal(10,2) [not null]
 
   Indexes {
     (order_id, product_id) [type: nonclustered, name: "idx_order_items_order_product"]
@@ -104,7 +117,7 @@ Table "dbo"."orders" {
   "order_id" int [pk, not null, increment]
   "shipping_address" text [not null]
   "status" dbo.chk_status [default: 'pending']
-  "total_amount" decimal [not null]
+  "total_amount" decimal(12,2) [not null]
   "user_id" int [not null]
 
   Indexes {
@@ -118,7 +131,7 @@ Table "dbo"."products" {
   "description" text
   "is_available" bit [default: 1]
   "name" varchar [not null]
-  "price" decimal [not null]
+  "price" decimal(10,2) [not null]
   "product_id" int [pk, not null, increment]
   "stock_quantity" int [not null, default: 0]
   "updated_at" datetime2 [default: `getdate()`]
--- a/packages/dbml-cli/__test__/db2dbml/mssql/expect-out-files/schema.dbml
+++ b/packages/dbml-cli/__test__/db2dbml/mssql/expect-out-files/schema.dbml
@@ -13,18 +13,18 @@ Enum "dbo"."chk_gender" {
 }
 
 Table "dbo"."Authors" {
-  "AuthorID" int [pk, not null]
-  "AuthorName" nvarchar [unique]
-  "BirthYear" int [unique]
-  "NationalityID" int [pk, not null]
+  "AuthorID" int(10) [pk, not null]
+  "AuthorName" nvarchar(100) [unique]
+  "BirthYear" int(10) [unique]
+  "NationalityID" int(10) [pk, not null]
 }
 
 Table "dbo"."Books" {
-  "AuthorID" int [pk, not null]
-  "BookID" int [pk, not null]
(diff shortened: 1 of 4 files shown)


packages/dbml-connector/src/connectors/mssqlConnector.ts
```diff
@@ -49,19 +49,51 @@ const getValidatedClient = async (connection: string): Promise<sql.ConnectionPoo
...
   if (MSSQL_DATE_TYPES.includes(data_type)) {
     return data_type;
   }
+
+  // timestamp is a synonym of rowversion
+  // microsoft recommends that we should not use timestamp, so we will convert it back to rowversion
+  // https://learn.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver15
+  if (data_type === 'timestamp') {
+    return 'rowversion';
+  }
```
huyphung1602, line 62: I don't think so, we should use the same `data_type` that match exactly the user's database nha.

packages/dbml-connector/src/connectors/mssqlConnector.ts
```diff
@@ -229,7 +261,7 @@ const generateTablesFieldsAndEnums = async (client: sql.ConnectionPool, schemas:
       constraints_with_row_number
     WHERE
       rn = 1
-      ${buildSchemaQuery('table_schema', schemas, 'WHERE')}
+      ${buildSchemaQuery('table_schema', schemas)}
```
huyphung1602, line 264: Why do you need to change this em? I believe the old one is correct 🤔

packages/dbml-connector/src/connectors/mssqlConnector.ts
```diff
@@ -49,19 +49,51 @@ const getValidatedClient = async (connection: string): Promise<sql.ConnectionPoo
...
+  // character_maximum_length is the lenght in bytes
+  // nchar and nvarchar store Unicode characters, each character needs 2 bytes
+  // so we have to divide it by 2 to get the correct maximum lenght in character.
+  if (character_maximum_length > 0) {
+    const maximum_length_in_character = (data_type === 'nchar' || data_type === 'nvarchar')
+      ? character_maximum_length / 2
+      : character_maximum_length;
+
+    return `${data_type}(${maximum_length_in_character})`;
+  }
```
huyphung1602, line 95: Looks good, please also add the url to the mssql documentation related to this re-calculation em.

packages/dbml-connector/src/connectors/mssqlConnector.ts
```diff
@@ -49,19 +49,51 @@ const getValidatedClient = async (connection: string): Promise<sql.ConnectionPoo
...
     return data_type;
   }
-  if (numeric_precision && numeric_scale && default_type === 'number') {
-    return `${data_type}(${numeric_precision},${numeric_scale})`;
+  // if precision != 0 => numeric-based column
+  if (numeric_precision) {
+    return numeric_scale
+      ? `${data_type}(${numeric_precision},${numeric_scale})`
+      : `${data_type}(${numeric_precision})`;
   }
```
huyphung1602, line 73: How about date/time data type em?

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyleminh01: No anh, the previous caused error since the generated query looks like this:
```
 WHERE
      rn = 1
      WHERE table_schema IN ('dbo')
```

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyleminh01: got it anh, I will return `timestamp` instead, and notice that we cannot use syntax `timestamp(n)`

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyleminh01: Let me check again a, I forgot it

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyphung1602: aw, i see. Ok em

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyphung1602: sure, we do not support `datetime_type(n)`

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyleminh01: Added a link in https://github.com/holistics/dbml/pull/645/commits/74b7e23ae137b61b30ed2ffdb2bbb5f9c727ad14

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyleminh01: For date/time, datetime2 and datetimeoffset support custom fractional seconds precision, see https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql?view=sql-server-ver15

So I fetch the numeric scale for them.

Fixed in commit https://github.com/holistics/dbml/pull/645/commits/0b972fc75d2280aff819329ee5293c698b361379

packages/dbml-cli/__test__/db2dbml/mssql/expect-out-files/schema.dbml
```diff
@@ -13,55 +13,70 @@ Enum "dbo"."chk_gender" {
...
+  "BookID" int(10) [pk, not null]
+  "ISBN" nvarchar(20) [unique]
+  "NationalityID" int(10)
+  "Title" nvarchar(200)
 }
 
 Table "dbo"."DatetimeTypes" {
   "DATECol" date [default: `getdate()`]
-  "DATETIME2Col" datetime2 [default: `sysdatetime()`]
+  "DATETIME2Col" datetime2(7) [default: `sysdatetime()`]
```
huyphung1602, line 32: We will not want to show the users the length of datetime types. Please help me to fix it nha.

huyleminh01: I reverted to this [commit](https://github.com/holistics/dbml/pull/645/commits/2310156d58c53e8fcf57974b58db9ecf69f9450d) nha a

huyphung1602 (approved):

PR #647, opened 2024-10-14
Fix model-exporter don't export sticky notes, table group notes and color

thonx-holistics:
## Summary
* Add a method to export sticky notes.
* Update the method that export table groups to also export table group notes and colors.

## Issue
(issue link here)

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/src/export/DbmlExporter.js
+++ b/packages/dbml-core/src/export/DbmlExporter.js
@@ -304,8 +304,7 @@ class DbmlExporter {
       const stickyNote = `Note ${note.name} {\n${escapedContent}\n}\n`;
 
       // Add a blank line between note elements
-      if (result) return result + '\n' + stickyNote;
-      return result + stickyNote;
+      return result ? result + '\n' + stickyNote : stickyNote;
     }, '');
   }
 
--- a/packages/dbml-core/src/export/DbmlExporter.js
+++ b/packages/dbml-core/src/export/DbmlExporter.js
@@ -1,4 +1,4 @@
-import _ from 'lodash';
+import { isEmpty, forIn } from 'lodash';
 import { shouldPrintSchema } from './utils';
 import { DEFAULT_SCHEMA_NAME } from '../model_structure/config';
 
@@ -185,19 +185,22 @@ class DbmlExporter {
     const tableStrs = tableContentArr.map((tableContent) => {
       const table = model.tables[tableContent.tableId];
       const schema = model.schemas[table.schemaId];
-      const tableSettingStr = this.getTableSettings(table);
+      const tableSettingStr = DbmlExporter.getTableSettings(table);
+      // Include schema name if needed
+      let tableName = `"${table.name}"`;
+      if (shouldPrintSchema(schema, model)) tableName = `"${schema.name}"."${table.name}"`;
+
+      const fieldStr = tableContent.fieldContents.map(field => `  ${field}\n`).join('');
 
       let indexStr = '';
-      if (!_.isEmpty(tableContent.indexContents)) {
-        indexStr = `\n  Indexes {\n${tableContent.indexContents.map(indexLine => `    ${indexLine}`).join('\n')}\n  }`;
+      if (!isEmpty(tableContent.indexContents)) {
+        const indexBody = tableContent.indexContents.map(indexLine => `    ${indexLine}\n`).join('');
+        indexStr = `\n  Indexes {\n${indexBody}  }\n`;
       }
-      const tableNote = table.note ? `  Note: ${DbmlExporter.escapeNote(table.note)}\n`: '';
 
-      const tableStr = `Table ${shouldPrintSchema(schema, model)
-        ? `"${schema.name}".` : ''}"${table.name}"${tableSettingStr} {\n${
-        tableContent.fieldContents.map(line => `  ${line}`).join('\n')}\n${indexStr ? `${indexStr}\n` : ''}${tableNote}}\n`;
+      const noteStr = table.note ? `  Note: ${DbmlExporter.escapeNote(table.note)}\n`: '';
 
-      return tableStr;
+      return `Table ${tableName}${tableSettingStr} {\n${fieldStr}${indexStr}${noteStr}}\n`;
     });
 
     return tableStrs.length ? tableStrs.join('\n') : '';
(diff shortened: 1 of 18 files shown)


packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -263,59 +266,67 @@ class DbmlExporter {
...
+      const groupSettingStr = DbmlExporter.getTableGroupSettings(group);
 
-      return `TableGroup ${shouldPrintSchema(groupSchema, model)
-        ? `"${groupSchema.name}".` : ''}"${group.name}" {\n${
-        group.tableIds.map(tableId => {
+      const groupNote = group.note ? `  Note: ${DbmlExporter.escapeNote(group.note)}\n`: '';
+      const groupSchemaName = shouldPrintSchema(groupSchema, model) ? `"${groupSchema.name}".` : '';
+      const groupName = `${groupSchemaName}"${group.name}"`;
+
+      const tableStr = group.tableIds
```
huyphung1602, line 287: `tableList` is a better name here. It avoid making confusion with the `tableStrs` above which store the table definitions.

huyphung1602: Or it should be: `tableNames` instead since it only a list of table names.

packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -263,59 +266,67 @@ class DbmlExporter {
...
     return tableGroupStrs.length ? tableGroupStrs.join('\n') : '';
   }
 
+  static exportStickyNotes (model) {
+    const noteStrs = [];
+    forIn(model.notes, (note) => {
+      const escapedContent = `  ${DbmlExporter.escapeNote(note.content)}`;
+      noteStrs.push(`Note ${note.name} {\n${escapedContent}\n}\n`);
+    });
+    return noteStrs.join('\n');
```
huyphung1602, line 308: Instead of writing the code like this. Use can use `reduce` from lodash to create the sticky notes string.

packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -263,59 +266,67 @@ class DbmlExporter {
...
-
-      if (!_.isEmpty(tableIds)) {
-        if (hasBlockAbove) res += '\n';
-        res += DbmlExporter.exportTables(tableIds, model);
-        hasBlockAbove = true;
-      }
+      if (!isEmpty(enumIds)) elementStrs.push(DbmlExporter.exportEnums(enumIds, model));
+      if (!isEmpty(tableIds)) elementStrs.push(DbmlExporter.exportTables(tableIds, model));
+      if (!isEmpty(tableGroupIds)) elementStrs.push(DbmlExporter.exportTableGroups(tableGroupIds, model));
+      if (!isEmpty(refIds)) elementStrs.push(DbmlExporter.exportRefs(refIds, model));
```
huyphung1602, line 323: The logic here appears to be different from the previous one. Why do we need to take care of these legacy codes? Do we have tests to ensure it won't cause any issues? cc @NQPhuc @thonx-holistics

packages/dbml-core/src/export/DbmlExporter.js
huyphung1602: It would be something like this. Please re-check.

```javascript
import { reduce } from 'lodash';

static exportStickyNotes (model) {
  return reduce(model.notes, (stickyNotes, note) => {
    const escapedContent = ` ${DbmlExporter.escapeNote(note.content)}`;
    const stickyNote =  `Note ${note.name} {\n${escapedContent}\n}\n`;
    stickyNotes += `${stickyNote}\n`;
    return stickyNotes;
  }, '');
};
```

packages/dbml-core/src/export/DbmlExporter.js
thonx-holistics: I updated it as `tableNames`

packages/dbml-core/src/export/DbmlExporter.js
thonx-holistics: I updated the code using lodash's `reduce` nha a

packages/dbml-core/src/export/DbmlExporter.js
NQPhuc: We do have some tests for `table, enum, ref` under `dbml-core/__test__/exporter/model_exporter/dbml_exporter`. Please help add more test for `table group, sticky note` @thonx-holistics

thonx-holistics: The logic should stay the same. Previously, before adding the DBML content of some elements to the result, the function will check if there is already an element in the DBML content to add a newline to separate them. Ultimately, the purpose of the old code is to separate DBML elements with newlines for clarity.

That's why I refactored it using the `join` method instead, to make the function more readable.

There are some unit tests to cover this, but I think they are not enough, so I'm writing some more tests.

NQPhuc (commented):
Please help add tag and update your PR description

packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -185,19 +185,22 @@ class DbmlExporter {
...
       }
-      const tableNote = table.note ? `  Note: ${DbmlExporter.escapeNote(table.note)}\n`: '';
 
-      const tableStr = `Table ${shouldPrintSchema(schema, model)
-        ? `"${schema.name}".` : ''}"${table.name}"${tableSettingStr} {\n${
-        tableContent.fieldContents.map(line => `  ${line}`).join('\n')}\n${indexStr ? `${indexStr}\n` : ''}${tableNote}}\n`;
+      const noteStr = table.note ? `  Note: ${DbmlExporter.escapeNote(table.note)}\n`: '';
 
-      return tableStr;
+      return `Table ${tableName}${tableSettingStr} {\n${fieldStr}${indexStr}${noteStr}}\n`;
```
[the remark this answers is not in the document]
thonx-holistics: I take this as example for implementing the export methods for table groups and sticky notes, but I find it kind of messy so I did a little refactor here.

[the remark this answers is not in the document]
NQPhuc: In that case, please make sure you have enough tests to account for all the table settings, fields settings and note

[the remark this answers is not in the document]
thonx-holistics: ok a, I will add more tests to cover those.

packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -263,59 +266,68 @@ class DbmlExporter {
...
   }
 
+  static exportStickyNotes (model) {
+    return reduce(model.notes, (result, note) => {
+      const escapedContent = `  ${DbmlExporter.escapeNote(note.content)}`;
+      const stickyNote = `Note ${note.name} {\n${escapedContent}\n}\n`;
+
+      // Add a blank line between note elements
+      if (result) return result + '\n' + stickyNote;
+      return result + stickyNote;
```
nguyenalter, line 308: - Can we just `return stickyNote` in the line 308?
- How about merging them into 1 line:
  ```js
  return result ? result + '\n' + stickyNote : stickyNote;
  ```

thonx-holistics: Thanks for the suggestion a, I updated the code.

nguyenalter (approved):

NQPhuc (approved):

PR #657, opened 2024-11-08
PR: Bug Fix #655 - DB2DBML - SQL Server - No description on Table

### LINKED ISSUE #655: DB2DBML - SQL Server - No description on Table
Before using the automatic generation of the DBML file for SQL server, I had queries allowing us to generate our file in DBML format.
In these scripts, I retrieved the description table, which was very useful when viewing tables in dbdocs
Exemple : 
```
Table "dbo"."B_BATCH_PLANIF" {
  "BBPL_ANNEE" numeric(14,2)
  "BBPL_AV_ANNEE" nvarchar(60) [note: 'Mode CRON : Annee(s)']
  "BBPL_AV_HEURE" nvarchar(10) [note: 'Mode CRON : Heures']
 Note: 'Table des planification du batch' 
}
```

This informations are stock on sys.extended_properties where sys.extended_properties.name = MS_Description_Table 

Other things :  With the actual script we do not respect the order of the columns. Is it possible to keep the same order as the initial script ? 
These information are very usefull particulary on our schema wich contains 742 tables and over 16 000 fields :-) 

Thanks a lot

MisterGeek:
Correct table name comment ( DB2DBML - SQL Server - No description on Table #655 )

## Summary
Correct the sqlRequest

## Issue
[https://github.com/holistics/dbml/issues/655]

## Checklist

- [x] Code Review


huyleminh01:
Hi @MisterGeek ,

Sorry for the late review.

I've updated the PR and added some fixes to your work.

The reason why I fetch descriptions with the pattern `%description%` is

- SQL Server Management Studio and Azure Studio use `MS_Description` to define the object description, so it should be the common way.
- Your definition is good, though, so just extend the matching case; it's good for now.

xuantho573 (approved):

PR #679, opened 2025-02-19
Fix: Unset not_null if null or not_null setting is not specified

nguyenalter:
## Summary
Fix: Unset not_null if null or not_null setting is not specified

## Issue
(issue link here)

- Set `not_null` to `undefined` if neither `null` nor `not null` configuration is specified
- `not_null` will only have a boolean value if there is `null` or `not null` in the field setting

## Checklist

- [ ] Code Review

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
       mssql:
-        image: mcr.microsoft.com/mssql/server:2019-latest
+        image: mcr.microsoft.com/mssql/server:2022-latest
         env:
           ACCEPT_EULA: Y
           SA_PASSWORD: dbml.123.123
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
       mssql:
-        image: mcr.microsoft.com/mssql/server:2019-CU30-ubuntu-20.04
+        image: mcr.microsoft.com/mssql/server:2019-latest
         env:
           ACCEPT_EULA: Y
           SA_PASSWORD: dbml.123.123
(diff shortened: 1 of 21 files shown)


.github/workflows/test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
       mssql:
-        image: mcr.microsoft.com/mssql/server:2019-latest
+        image: mcr.microsoft.com/mssql/server:2019-CU30-ubuntu-20.04
```
NQPhuc, line 34: Since the un-related tests is still failed, please revert this. We'll fix the tests in another PR

packages/dbml-parse/src/lib/interpreter/elementInterpreter/table.ts
```diff
@@ -141,14 +141,16 @@ export class TableInterpreter implements ElementInterpreter {
...
     const settings = field.args.slice(1);
     if (_.last(settings) instanceof ListExpressionNode) {
       const settingMap = aggregateSettingList(settings.pop() as ListExpressionNode).getValue();
       column.pk = !!settingMap['pk']?.length || !!settingMap['primary key']?.length;
       column.increment = !!settingMap['increment']?.length;
       column.unique = !!settingMap['unique']?.length;
-      column.not_null = !!settingMap['not null']?.length && true;
+
+      const hasNullSetting = !!settingMap['null']?.length || !!settingMap['not null']?.length;
+      column.not_null = hasNullSetting ? !!settingMap['not null']?.length : undefined;
```
NQPhuc, line 153: I think this is clearer
```suggestion
      column.not_null = !!settingMap['not null']?.length ? true : !!settingMap['null']?.length ? false : undefined;
```

.github/workflows/test.yml
nguyenalter: ye, I reverted in c4e0cf6

packages/dbml-parse/src/lib/interpreter/elementInterpreter/table.ts
nguyenalter: Ok, I changed to nested ternary in 45b8a91

NQPhuc (approved):

PR #713, opened 2025-04-23
fix(antlr.pg): filter null table when importing postgresql

NQPhuc:
## Summary
* Filter null table when importing postgresql
* Add dummy function charIsLetter to pg lexer base to avoid lexing error when handle weird character (for example: curly quotation)

## Issue
(issue link here)

## Checklist

- [ ] Code Review


packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
```diff
@@ -65,6 +64,9 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
     if (ctx.createstmt()) {
       const table = ctx.createstmt().accept(this);
 
+      // filter out null table that can cause error in model_structure stage
+      if (!table) return undefined;
+
       return this.data.tables.push(table);
```
xuantho573, line 70: I suggest using 
```
if (table) this.data.tables.push(table);
return;
```
as it behave the same and remove unnecessary `return undefined` blocks after this

packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
```diff
@@ -74,29 +76,31 @@ export default class PostgresASTGen extends PostgreSQLParserVisitor {
       const { tableName, schemaName } = indexStmt.pathName;
 
       const table = findTable(this.data.tables, schemaName, tableName);
-      if (!table) return;
+      if (!table) return undefined;
       return table.indexes.push(indexStmt.index);
```
xuantho573, line 80: Same as above

packages/dbml-core/src/parse/ANTLR/ASTGeneration/postgres/PostgresASTGen.js
NQPhuc: There's a lint error that said all branch should return somethin. That's why I'm adding `undefined`. It won't affect anything.

NQPhuc: I update this and fix all lint error in this file in https://github.com/holistics/dbml/pull/713/commits/45d8f71f40a8ba3d4ede6874a71c644036a9f99d

xuantho573 (approved):

PR #714, opened 2025-04-24
Fix flaky db2dbml/mssql test cause by inconsistent table ordering

NQPhuc:
## Summary
* Order table with same create_date by object_id instead of table create date in mssql connector
* Upgrade tests ci node version to 22

## Issue
(issue link here)

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-connector/src/connectors/mssqlConnector.ts
+++ b/packages/dbml-connector/src/connectors/mssqlConnector.ts
@@ -215,6 +215,7 @@ const generateTablesFieldsAndEnums = async (client: sql.ConnectionPool, schemas:
   const tablesAndFieldsSql = `
     WITH tables_and_fields AS (
       SELECT
+        t.object_id AS table_id,
         s.name AS table_schema,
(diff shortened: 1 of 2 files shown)


packages/dbml-connector/src/connectors/mssqlConnector.ts
```diff
@@ -307,6 +308,7 @@ const generateTablesFieldsAndEnums = async (client: sql.ConnectionPool, schemas:
     ORDER BY
       table_schema,
       table_create_date,
```
huyleminh01, line 310: you can remove this condition since the `table_id` is enough

NQPhuc: My theory is that some table create_date are the same, and we only need to sort those via id. I still haven't sure if id object is always increment

huyleminh01: ye it's ok

huyleminh01 (approved):

PR #720, opened 2025-04-28
Fix not handling column type with arguments in TablePartial

xuantho573:
## Summary
* Fix TablePartial not handling column type that have arguments (fix DBX-5987)
image
  * `function processColumnType` in TablePartial extracted the name from the original typeNode, which is `CallExpressionNode` for column type with arguments
  * It should have extracted the callee of the call expression instead

* Fix ESLint for
  * `interpreter/elementInterpreter/table.ts`
  * `interpreter/elementInterpreter/tablePartial.ts`
  * `interpreter/utils.ts`

## Issue
(issue link here)

* Moved `function processDefaultValue` and `function processColumnType` from `table.ts` and `tablePartial.ts` to `utils.ts` as those 2 files use the same helper functions

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-parse/src/lib/interpreter/elementInterpreter/table.ts
+++ b/packages/dbml-parse/src/lib/interpreter/elementInterpreter/table.ts
@@ -1,27 +1,25 @@
 import _ from 'lodash';
 import {
-  Column, ColumnType, ElementInterpreter, Index,
-  InlineRef, InterpreterDatabase, Table, TablePartialInjection,
+  Column, ElementInterpreter, Index, InlineRef,
+  InterpreterDatabase, Table, TablePartialInjection,
 } from '../types';
 import {
-  ArrayNode, AttributeNode, BlockExpressionNode, CallExpressionNode,
-  ElementDeclarationNode, FunctionApplicationNode, FunctionExpressionNode, ListExpressionNode,
-  PartialInjectionNode, PrefixExpressionNode, SyntaxNode,
+  AttributeNode, BlockExpressionNode, CallExpressionNode, ElementDeclarationNode,
+  FunctionApplicationNode, ListExpressionNode, PartialInjectionNode, PrefixExpressionNode,
+  SyntaxNode,
 } from '../../parser/nodes';
 import {
   extractColor, extractElementName, getColumnSymbolsOfRefOperand, getMultiplicities,
   getRefId, getTokenPosition, isSameEndpoint, normalizeNoteContent,
+  processColumnType, processDefaultValue,
 } from '../utils';
 import {
   destructureComplexVariable, destructureIndexNode, extractQuotedStringToken, extractVarNameFromPrimaryVariable,
   extractVariableFromExpression,
 } from '../../analyzer/utils';
 import { CompileError, CompileErrorCode } from '../../errors';
-import { aggregateSettingList, isExpressionANumber } from '../../analyzer/validator/utils';
-import { isExpressionAQuotedString, isExpressionAnIdentifierNode } from '../../parser/utils';
-import { NUMERIC_LITERAL_PREFIX } from '../../../constants';
+import { aggregateSettingList } from '../../analyzer/validator/utils';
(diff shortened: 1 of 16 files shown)


packages/dbml-parse/src/lib/interpreter/elementInterpreter/table.ts
```diff
@@ -110,12 +108,14 @@ export class TableInterpreter implements ElementInterpreter {
     return [];
   }
 
-  private interpretSettingList(settings?: ListExpressionNode): CompileError[] {
+  private interpretSettingList (settings?: ListExpressionNode): CompileError[] {
     const settingMap = aggregateSettingList(settings).getValue();
 
-    this.table.headerColor = settingMap['headercolor']?.length ? extractColor(settingMap['headercolor']?.at(0)?.value as any) : undefined;
+    this.table.headerColor = settingMap.headercolor?.length
+      ? extractColor(settingMap.headercolor?.at(0)?.value as any)
```
huyphung1602, line 115: Which did you change from settingMap['key'Name] object type to `settingMap.keyName`? Should we use `get` from lodash to support both cases? cc @nguyenalter

huyphung1602:
Can you be more collaborate inside the PR's description? For example:
This PR will fix the issue related to ... Example:
- Example 1
- Example 2

The root cause of this issue is: [write the explanation of the root cause] => try to describe which is causing the root cause.

Then, tell your high-level solution in this PR.

Moreover, I see that you fix some code that are not related to the issue. Isn't it?

packages/dbml-parse/src/lib/interpreter/elementInterpreter/table.ts
xuantho573: I updated the syntax to use `settingMap[SettingName.keyName]` instead. Since `SettingName` is an Enum with lowercase values that match currently supported setting names in dbml, the new syntax will behave the same as the old syntax nha a.
Updated in https://github.com/holistics/dbml/pull/720/commits/b8544e7304e04f32a3ef3d31e39ee1688c5e803b

huyphung1602: Did we already have tests to cover this?

xuantho573: Yes anh, we have tests where tables can contain either `note` or `headercolor`, or both, or none of the settings. There are also tests where these settings are in different casing (interpreter test `header_color_table`, ...).
For columns, these tests have many combinations of settings too, including both single-word settings and multiple-word settings

huyphung1602 (approved):

packages/dbml-parse/src/lib/interpreter/elementInterpreter/table.ts
```diff
@@ -187,23 +187,23 @@ export class TableInterpreter implements ElementInterpreter {
...
-      column.unique = !!settingMap['unique']?.length;
-      column.not_null = !!settingMap['not null']?.length
+      column.pk = !!settingMap[SettingName.PK]?.length || !!settingMap[SettingName.PKey]?.length;
+      column.increment = !!settingMap[SettingName.Increment]?.length;
+      column.unique = !!settingMap[SettingName.Unique]?.length;
+      // eslint-disable-next-line no-nested-ternary
+      column.not_null = settingMap[SettingName.NotNull]?.length
         ? true
-        : !!settingMap['null']?.length
+        : settingMap[SettingName.Null]?.length
```
NQPhuc, line 196: why remove `!!`?

xuantho573: ESLint raise the `Redundant double negation` error, and the ternary op use truthy and falsy conditions, so we don't have to convert the conditions to actual booleans here anh

NQPhuc (commented):
@xuantho573 Can you help add more test in `dbml-core` to make sure we don't choke on special expression in `column type` and `default value` of table_partial?

NQPhuc (approved):

PR #727, opened 2025-05-05
Fix table must define column when have partials with columns

xuantho573:
## Summary
* Allow table to not define column when having table partials with columns inside the body (resolve DBX-5994)
* Please include screenshots whenever possible (important).

## Issue
(issue link here)

## Checklist

- [ ] Code Review


packages/dbml-parse/src/lib/interpreter/elementInterpreter/table.ts
```diff
@@ -163,11 +164,28 @@ export class TableInterpreter implements ElementInterpreter {
...
+    const interpretFieldErrors = fields.flatMap((field, order) => {
       return field instanceof FunctionApplicationNode
         ? this.interpretColumn(field)
         : this.interpretInjection(field, order);
     });
+
+    return [
+      ...columnCountErrors,
+      ...interpretFieldErrors,
+    ];
```
NQPhuc, line 188: Can you explain why we add the error here?

xuantho573: I'm not sure if your question is
1. `why add the error at this method of the interpreter?` or
2. `why add the error at the interpreting step?`
, so I will answer both:
1. The error is for counting the number of columns, and columns are fields, so this is the scope for the validation. Also, the old check was at the `validateFields` method of the TableValidator.
2. We don't know the total number of columns in a table until after we resolve the injections, therefore we have 2 places to check for this: binding step and interpreting step. The structure of the Binder is hard to change for specific element type because would have to add custom check for element type, while in the Interpreter we already have interpreter for each element type, so it's easier to check here

NQPhuc: What I mean is why do we need this check while there're already a duplicate check is module structure. But it's fine to keep it here for now

PR #728, opened 2025-05-05
Feat: allow unlimited prefix expression in column type with argument and default value

NQPhuc:
## Summary
Fix DBX-5979

* Update the lexer to parse chained operator tokens separately unless they are multi-chars operators
Example:
```
Input:  != +-+-
Before: token(!=) token(+-+-)
Now: token(!=) token(+) token(-) token(+) token(-)
```

* This allow the parser to nest prefix expression of chained operator into proper tree
Example:
```
Input:  -+-3
Before: PrefixExpression("-+-", 3)
Now: PrefixExpression("-", PrefixExpression("+", PrefixExpression("-", 3)))
```

* Update validator and interpreter to traverse nested PrefixExpression. Now the following syntax is valid for both table and table partial (since they share the same validate and interpret methods)
```
Table a {
  id int(-1) [default: -+-3]
}
```

## Issue
(issue link here)

## Checklist

- [ ] Code Review


packages/dbml-parse/src/lib/analyzer/validator/utils.ts
```diff
@@ -197,9 +199,13 @@ export function isValidDefaultValue(value?: SyntaxNode): boolean {
   return false;
 }
 
-export function isExpressionANumber(value?: SyntaxNode): value is PrimaryExpressionNode & {
-  expression: LiteralNode & { literal: { kind: SyntaxTokenKind.NUMERIC_LITERAL } };
+export function isExpressionANumber (value?: SyntaxNode): value is PrimaryExpressionNode | PrefixExpressionNode & {
+  op?: SyntaxToken, expression: LiteralNode & { literal: { kind: SyntaxTokenKind.NUMERIC_LITERAL } };
```
xuantho573, line 203: The type here results in `PrimaryExpressionNode | (PrefixExpressionNode & {...})`, which is wrong for our class declarations (`expression` field of `PrefixExpressionNode` is of type `NormalExpressionNode`, which does not include `LiteralNode`). I suggest changing it to
```
(
  PrefixExpressionNode
  & { expression: PrimaryExpressionNode | PrefixExpressionNode }
)
| (... the old type)
```

NQPhuc: I changed the return type to boolean in https://github.com/holistics/dbml/pull/728/commits/304cccadbd9653a4867398b0b7ac7abf9ba13dc6

Issue #286, opened 2022-09-02
[Feature Request] CLI supports SQLite 3


bxk-sonavex:

zenomeki:
I've got an upcoming open source project and this would come in really handy for it!

Issue #14, opened 2019-08-10
Add SQLite support


rasmusagren:
It would be great if you could add SQLite export. Thanks!

phuongduyphan:
Hi @rasmusagren ,

Thank you for your feedback!

At the moment, we're preparing to open source the dbml core package to the community to get more support for different database languages. SQLite will be supported soon !

JohnMarsing:
**SQL Server** support would be great as well.  I've used the SSMS database diagram tool  from Microsoft in the past and tolerated the **numerous times it crashed**, but now it has gotten even worse, so looking for other solutions.

michaeltryby:
@phuongduyphan What is your time frame for adding support for SQLite? We could use this now. Thanks!

edridgedsouza:
Has there been any update? For now it seems that it can be achieved by a third-party [tool](https://stackoverflow.com/questions/3890518/convert-mysql-to-sqlite), though it'd be nice if it were integrated natively

GoodiesHQ:
I'm very interested in this! Has any progress been made? It would seem SQLite is among the top 5 most popular databases in modern times, incredibly easy to set up and use for small projects.

PR #752, opened 2025-09-09
feat(dbml/core): adds support for exporting to sqlite

danielgary:
## Summary
Implemented SQLite exporter with many-to-many relations and proper index exporting support.  Included tests.

## Issue
#14 
#286

* Added sqlite as an export option for the cli
* Implented SQLiteExporter class

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-core/src/export/SqliteExporter.js
+++ b/packages/dbml-core/src/export/SqliteExporter.js
@@ -11,41 +11,13 @@ function escapeSingleQuotes(s) {
   return String(s || '').replace(/'/g, "''");
 }
 
-const SQLITE_TYPE_MAP = (() => {
-  const int = new Set(['SMALLINT', 'INT2', 'INTEGER', 'INT', 'INT4', 'BIGINT', 'INT8',
-    'SMALLSERIAL', 'SERIAL', 'BIGSERIAL']);
-  const real = new Set(['REAL', 'FLOAT', 'DOUBLE', 'DOUBLE PRECISION', 'NUMERIC(.*)']);
-  const num = new Set(['DECIMAL', 'NUMERIC']);
-  const text = new Set(['CHAR', 'CHARACTER', 'NCHAR', 'NVARCHAR', 'VARCHAR', 'CHARACTER VARYING', 'TEXT', 'NAME', 'BPCHAR', 'UUID', 'XML', 'JSON', 'JSONB', 'INET', 'CIDR', 'MACADDR', 'MACADDR8']);
-  const blob = new Set(['BYTEA', 'BLOB']);
-  const date = new Set(['DATE', 'TIME', 'TIMESTAMP', 'TIMESTAMP WITH TIME ZONE', 'TIMESTAMP WITHOUT TIME ZONE', 'TIME WITH TIME ZONE', 'TIME WITHOUT TIME ZONE', 'INTERVAL']);
-  const bool = new Set(['BOOLEAN', 'BOOL']);
-  return { int, real, num, text, blob, date, bool };
-})();
+// Minimal boolean type detection - only used to add CHECK (field IN (0,1)) constraints
+// for fields that should be boolean in nature
+const BOOLEAN_TYPES = new Set(['BOOLEAN', 'BOOL']);
 
-function normalizeTypeName(tn) {
-  const t = (tn || '').trim().replace(/\s+/g, ' ').toUpperCase();
-  return t.replace(/\(.*\)$/, '');
-}
-
-function mapTypeToSQLite(originalTypeName) {
-  const t = normalizeTypeName(originalTypeName);
-
-  if (SQLITE_TYPE_MAP.int.has(t)) return 'INTEGER';
-  if (SQLITE_TYPE_MAP.bool.has(t)) return 'INTEGER';
-  if (SQLITE_TYPE_MAP.blob.has(t)) return 'BLOB';
-  if (SQLITE_TYPE_MAP.text.has(t)) return 'TEXT';
-  if (SQLITE_TYPE_MAP.date.has(t)) return 'TEXT';
-  if (SQLITE_TYPE_MAP.num.has(t)) return 'NUMERIC';
-  if (t === 'REAL' || t === 'FLOAT' || t === 'DOUBLE' || t === 'DOUBLE PRECISION') return 'REAL';
-
-  if (/INT/i.test(t)) return 'INTEGER';
-  if (/CHAR|CLOB|TEXT/i.test(t)) return 'TEXT';
-  if (/BLOB/i.test(t)) return 'BLOB';
-  if (/REAL|FLOA|DOUB/i.test(t)) return 'REAL';
-  if (/NUM|DEC|MONEY/i.test(t)) return 'NUMERIC';
-
-  return 'TEXT';
+function isBooleanType(typeName) {
+  const normalized = (typeName || '').trim().toUpperCase();
+  return BOOLEAN_TYPES.has(normalized);
 }
 
 function topoSortTables(allTableIds, dependencyEdges) {
@@ -140,10 +112,10 @@ class SqliteExporter {
         const key2s = [...secondTableFieldsMap.keys()].join('","');
 
         firstTableFieldsMap.forEach((fieldType, fieldName) => {
-          line += `  "${fieldName}" ${mapTypeToSQLite(fieldType)},\n`;
+          line += `  "${fieldName}" ${fieldType},\n`;
         });
         secondTableFieldsMap.forEach((fieldType, fieldName) => {
-          line += `  "${fieldName}" ${mapTypeToSQLite(fieldType)},\n`;
+          line += `  "${fieldName}" ${fieldType},\n`;
         });
 
         line += `  PRIMARY KEY ("${key1s}","${key2s}"),\n`;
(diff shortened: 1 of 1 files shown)


danielgary:
Just curious if I missed something in this PR?  I'm happy to fix if something is missing.  Can someone on this project provide some guidance?

huydo862003:
Hey @danielgary, sorry for the late reply. Thanks for the contribution! We'll definitely take a look at your PR when we have time. Thank you for your patience.

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,367 @@
...
+  const real = new Set(['REAL', 'FLOAT', 'DOUBLE', 'DOUBLE PRECISION', 'NUMERIC(.*)']);
+  const num = new Set(['DECIMAL', 'NUMERIC']);
+  const text = new Set(['CHAR', 'CHARACTER', 'NCHAR', 'NVARCHAR', 'VARCHAR', 'CHARACTER VARYING', 'TEXT', 'NAME', 'BPCHAR', 'UUID', 'XML', 'JSON', 'JSONB', 'INET', 'CIDR', 'MACADDR', 'MACADDR8']);
+  const blob = new Set(['BYTEA', 'BLOB']);
+  const date = new Set(['DATE', 'TIME', 'TIMESTAMP', 'TIMESTAMP WITH TIME ZONE', 'TIMESTAMP WITHOUT TIME ZONE', 'TIME WITH TIME ZONE', 'TIME WITHOUT TIME ZONE', 'INTERVAL']);
+  const bool = new Set(['BOOLEAN', 'BOOL']);
+  return { int, real, num, text, blob, date, bool };
+})();
+
+function normalizeTypeName(tn) {
```
huydo862003: Currently, this function performs the following:
* Remove white spaces
* Turn all characters to uppercase
* Remove `(.*)`
I think you should just [quote the type name in double quote](https://stackoverflow.com/questions/36338815/sqlite-column-name-quoting-feature-or-bug) instead (if it contains special characters).

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,367 @@
...
+  buildJunctionFields2,
+  buildNewTableName,
+} from './utils';
+import { shouldPrintSchemaName } from '../model_structure/utils';
+
+function escapeSingleQuotes(s) {
+  return String(s || '').replace(/'/g, "''");
+}
+
+const SQLITE_TYPE_MAP = (() => {
```
huydo862003: I think you can just assign:
```
SQLITE_TYPE_MAP = {
   int: ...
   real: ...
   ...
}
```

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,367 @@
...
+
+function normalizeTypeName(tn) {
+  const t = (tn || '').trim().replace(/\s+/g, ' ').toUpperCase();
+  return t.replace(/\(.*\)$/, '');
+}
+
+function mapTypeToSQLite(originalTypeName) {
+  const t = normalizeTypeName(originalTypeName);
+
+  if (SQLITE_TYPE_MAP.int.has(t)) return 'INTEGER';
```
huydo862003: So basically your trying to map DBML types to Sqlite data types. Can you add a comment that points to SQLite's documentation showing the following points:
* `INTEGER` is the only integral type in Sqlite. Boolean is represented as `INTEGER`. Do the same for the rest of the types.
* Are booleans, blobs, texts, dates, nums types the only types? Are there other kinds of types.

Suggestion: Maybe you should just keep the type "as is", instead of trying to map. We accept that the type may not exist in SQLite

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,367 @@
...
+  if (SQLITE_TYPE_MAP.bool.has(t)) return 'INTEGER';
+  if (SQLITE_TYPE_MAP.blob.has(t)) return 'BLOB';
+  if (SQLITE_TYPE_MAP.text.has(t)) return 'TEXT';
+  if (SQLITE_TYPE_MAP.date.has(t)) return 'TEXT';
+  if (SQLITE_TYPE_MAP.num.has(t)) return 'NUMERIC';
+  if (t === 'REAL' || t === 'FLOAT' || t === 'DOUBLE' || t === 'DOUBLE PRECISION') return 'REAL';
```
huydo862003: This seems inconsistent: Why not just test SQLITE_TYPE_MAP.real.has(t)?

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,367 @@
...
+
+  if (/INT/i.test(t)) return 'INTEGER';
```
huydo862003: These additional checks seem redundant and adhoc to me?

huydo862003:
Thank you for resolving all my comments, I'll continue reviewing when I have time!

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,338 @@
...
+
+      line += ` (${columnArr.join(',')})`;
+      line += ';\n';
+      return line;
+    });
+
+    return indexArr;
+  }
+
+  static export(model) {
```
huydo862003, line 304: Since dbml v5.0.0, we have supported check constraints in model_structure. The `model` parameter now can contain information about check constraints.

You can look at the definition here for reference:
- Table: https://github.com/holistics/dbml/blob/master/packages/dbml-core/src/model_structure/table.js
- Field: https://github.com/holistics/dbml/blob/master/packages/dbml-core/src/model_structure/field.js
- Check: https://github.com/holistics/dbml/blob/master/packages/dbml-core/src/model_structure/check.js

Please help us exporting Check constraints as well!

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,338 @@
...
+    const database = model.database['1'];
+
+    const enumMap = SqliteExporter.buildEnumMap(model);
+
+    const allRefIds = _.flatten(database.schemaIds.map(sid => model.schemas[sid].refIds || []));
+    const { fksByTableId, junctionCreates, dependencyEdges } = SqliteExporter.collectForeignKeysByTable(allRefIds, model);
+
+    const allTableIds = _.flatten(database.schemaIds.map(sid => model.schemas[sid].tableIds || []));
```
huydo862003, line 312: Can you use `flatMap` instead of `map` + `flatten`?

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,338 @@
...
+  static buildEnumMap(model) {
+    const map = new Map();
+    Object.keys(model.enums).forEach((enumId) => {
+      const e = model.enums[enumId];
+      const schema = model.schemas[e.schemaId];
+      const fq = `"${schema.name}"."${e.name}"`;
+      const local = `"${e.name}"`;
+      const vals = e.valueIds.map(id => model.enumValues[id].name);
+      map.set(fq, vals);
+      map.set(local, vals);
```
huydo862003, line 74: I wonder if this approach would work if there are two enums in two different schemas but have the same name? For example:

```
Enum S1.E {
}

Enum S2.E {
}
```

By performing `map.set(local, vals)`, `E` would be set two times and the latter would override.

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,338 @@
...
+
+    return indexArr;
+  }
+
+  static export(model) {
+    const database = model.database['1'];
+
+    const enumMap = SqliteExporter.buildEnumMap(model);
+
+    const allRefIds = _.flatten(database.schemaIds.map(sid => model.schemas[sid].refIds || []));
```
huydo862003, line 309: Can you use flatMap instead of map + flatten?

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,338 @@
...
+    const { fksByTableId, junctionCreates, dependencyEdges } = SqliteExporter.collectForeignKeysByTable(allRefIds, model);
+
+    const allTableIds = _.flatten(database.schemaIds.map(sid => model.schemas[sid].tableIds || []));
+    const orderedTableIds = topoSortTables(allTableIds, dependencyEdges);
+
+    const tableCreates = SqliteExporter.exportTables(orderedTableIds, model, enumMap, fksByTableId);
+
+    const allIndexIds = _.flatten(
```
huydo862003, line 317: Can you use flatMap instead of map + flatten?

packages/dbml-core/src/export/SqliteExporter.js
```diff
@@ -0,0 +1,338 @@
...
+      (database.schemaIds || []).map(sid => {
+        const tIds = model.schemas[sid].tableIds || [];
+        return _.flatten(tIds.map(tid => model.tables[tid].indexIds || []));
```
huydo862003, line 320: Can you use flatMap instead of map + flatten?

packages/dbml-core/__tests__/exporter/exporter.spec.js
huydo862003, line 1: The testcases are failing. Please help update the tests!
Also, can you verify that the following cases are covered in the test cases:
- Fields that have types in the form of `name(args)` such as `VARCHAR(255)`.
- Various index types: unique, primary, functional, multi-column (unique and primary).
- Unique constraints.
- Default
- Autoincrement.
- Various referential actions.
- Primary constraints.
- Check constraints.
- Enums
- Multi-column, many-to-many, one-to-one, one-to-many, zero-to-one, zero-to-many Refs.
- Not null constraints
- Column types that have spaces in their names.
- Table and column names that contain spaces in them
You can look up the input for the test cases by reusing the other input files in the other exporters.

Issue #591, opened 2024-07-17
Workaround for column un-supported settings doesn't work


gguy0406:
### Problem
The workaround for un-supported settings by adding the setting name into the column type name, such as id “bigint unsigned” [pk] doesn't work
### Example
Take an example DBML file:
```
Table Product {
  id integer [increment, pk]
  price "integer check (price > 0)" [not null]
  quantity "integer check(quantity > 0) [default: 0, not null]
}
```
When I convert that to Postgresql, it looks like:
```
CREATE TABLE "Product " (
  "id" INTEGER GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
  "price" "integer check (price > 0)" NOT NULL,
  "quantity" "integer check (quantity > 0)" NOT NULL DEFAULT 0,
);
```
Which then throwing error when attempting to execute query `pq: type "integer check (totalPrice > 0)" does not exist`

NQPhuc:
Hi,

Currently, there's no way to add check constraint and other unsupported settings so the the export SQL is correctly equivalent.

The work-around is only mean to make the dbml valid. Instead of putting the constraint in the column type, you can try to annotate the constraint in `field note`, like this:

```
Table Product {
  id integer [increment, pk]
  price integer [not null, note: "check (price > 0)"]
  quantity integer [default: 0, not null, note: "check (quantity > 0)"]
}
```

Which will export to Postgresql without error, but you'll have to add the check constraint back mannually.

 We'll consider support the CHECK constraint in the future as well, since it is a common problem.

huydo862003:
Check constraints have been supported in version ^v5.0.0: https://dbml.dbdiagram.io/docs/#check-definition

Issue #484, opened 2023-12-11
Add support for custom constraints 


iloudaros:
for example, how can I use dbml to give this description :

```
CREATE TABLE `MyTable` (
  `id` INT PRIMARY KEY AUTO_INCREMENT,
  `foreignKey1` INT,
  `foreignKey2` INT,
  CONSTRAINT `check_foreign_key` CHECK (`foreignKey1` IN (SELECT `id` FROM `OtherTable1`) OR `foreignKey2` IN (SELECT `id` FROM `OtherTable2`)))
);
```

NQPhuc:
Hi,

For the columns and table, you can define dbml like this (However, there's currently no equivalent dbml syntax for CHECK CONTRAINTS):
```
Table "MyTable" {
  "id" INT [pk, increment]
  "foreignKey1" INT
  "foreignKey2" INT
}
```

huydo862003:
Check constraints have been supported in dbml ^v5.0.0: https://dbml.dbdiagram.io/docs/#check-definition

Issue #68, opened 2020-02-08
Column Constraints?


officer-rosmarino:
How can I express constraints? For instance, I need to express that the value of column rating is between >0 and <=5? I see no documentation. Is this possible?

phuongduyphan:
Hi @officer-rosmarino ,

Currently we haven't supported to define column constraints as in your case.
For now, I think you can work-around by using column note syntax to express the constraints like this: 
```
Table movies {
  id int [pk]
  rating int [note: '>0 && <=5']
}
```

estan:
This is something I would need as well.

I also need to express table constraints, for example, to constrain the value in one column to be greater than that in another. Should I file a separate issue for table constraints?

djpeach:
So it seems there are two constraints that are currently not possible in DBML. That would be the `CHECK` constraint, discussed here, and a composite `UNIQUE` constraint with multiple columns. The composite `UNIQUE` constraint can be done with a unique index, and is functionally the same thing, but there are cases where the semantics of the Table definitions would be better served with a composite unique constraint instead. 

Off the top of my head a good syntax would be something like 

```
Table Foo {
  ...

  constraints {
    <constraint_type> [options]
  }
}
```

This does break convention with indexes, in that you usually do `(columns) [options]`. But holds with what the [docs say](https://www.dbml.org/docs/#syntax-consistency): curly brackets {}: grouping for indexes, **constraints** and table definitions. 

So for unique constraints: 
```
Table Foo {
  ...

  constraints {
    unique [name: 'UC_Col1_Col2', columns: (Col1, Col2)]
  }
}
```

And for check constraints:
```
Table Foo {
  ...

  constraints {
    unique [name: 'CC_Age_GTE_21', expression: 'Age >= 21']
  }
}
```

@officer-rosmarino would this do what you need, and @phuongduyphan would this be feasible and play well with what currently exists? If so, I would be willing to dig into this and attempt these two constraint types for the currently supported SQL database types.

Yokozuna59:
It would great if we've this feature! I'm aware that I could create a script that extract constraints from notes, but not clean nor efficient.

I think the recommended syntax by @djpeach looks great, but could you please provide some more complex examples? I might be able to help, although I'm new to `PEG.js` and `dbml`, but I'll try since it's an essential feature.

Yokozuna59:
Here is one:

```dbml
Table dog {
  id INTEGER
  name VARCHAR
  age INTEGER

  constraints {
[...]

PR #765, opened 2025-10-14
Feat: Support check constraint

huydo862003:
## Summary
* `dbml-parse`:
   * Support `constraint` setting in table columns and tablepartial columns: a column can have multiple constraints.
     ```
         Table User {
               balance int [constraint: `balance > 0`, constraint: `balance < 100000000`]
         }
     ```
   * Support `constraints` element in table and tablepartial elements: A table can contain multiple constraints and a constraint can be named or nameless.
     ```
        Table User {
               balance int
               constraints {
                   `balance >= 0` [name: "balance_not_negative"]
                   `balance < 1000000000`
               }
        }
     ```
   * Constraints in TablePartial will be injected into the Table that injects the partial. Constraint names injected this way will be modified properly to avoid duplicate constraint names.
   * Support autocompletion related to `constraints` and `constraint`.
     * When inside a column setting, `constraint` would be suggested.
     * When inside a Table element, `constraints` would be suggested.
     * When inside a constraint setting, `name` would be suggested.
   * Function expressions (\`expr\`) can now contain newlines to better support complex conditions. Example:
     ```
         Table inventory {
             stock int
             status string
             constraints {
                `CASE WHEN status = 'active' THEN stock
                            WHEN status = 'discontinued' THEN 0
                            ELSE 1
                  END >= 0`
             }
         }
     ```
  * Some minor bug fixes and refactors.
* `dbml-core`:
  * Support `constraints` in model structure.
  * Support `constraints` in all parsers & exporters.
    * Support column's check constraints.
    * Support table definition's check constraints.
    * Support ALTER TABLE ADD CONSTRAINT.
  * Remove `CONSTRAINT` as plsql_unreserved_keyword from Postgres ANTLR grammar.
  * No longer support converting CHECK constraints into DBML ENUMs in MSSQL parser.
* Not supported yet:
  * ALTER TABLE ALTER CONSTRAINT is not supported and is currently ignored when importing SQL. 
  * ALTER TABLE DROP CONSTRAINT is not supported and is currently ignored when importing SQL.
  * ALTER TABLE RENAME CONSTRAINT is not supported.
  * Duplicate constraint names are allowed. We currently do not perform duplicate constraint names check.
  
## Issue
#591 
#484
#68 

## Checklist

- [ ] Code Review

### DIFF
--- a/dbml-playground/package.json
+++ b/dbml-playground/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dbml/playground",
-  "version": "3.15.0-alpha.0",
+  "version": "3.14.1",
   "description": "Interactive playground for debugging and visualizing the DBML parser pipeline",
   "author": "Holistics <dev@holistics.io>",
   "license": "Apache-2.0",
@@ -24,7 +24,7 @@
     "format": "prettier --write src/"
   },
   "dependencies": {
-    "@dbml/parse": "^3.15.0-alpha.0",
+    "@dbml/parse": "^3.14.1",
     "lodash": "^4.17.21",
     "monaco-editor": "^0.52.2",
     "monaco-vim": "^0.4.2",
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "3.15.0-alpha.0",
+  "version": "3.14.1",
   "npmClient": "yarn",
   "$schema": "node_modules/lerna/schemas/lerna-schema.json"
 }
--- a/packages/dbml-cli/package.json
+++ b/packages/dbml-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dbml/cli",
-  "version": "3.15.0-alpha.0",
+  "version": "3.14.1",
   "description": "",
   "main": "lib/index.js",
   "license": "Apache-2.0",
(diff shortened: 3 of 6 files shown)


packages/dbml-core/src/parse/ANTLR/parsers/postgresql/PostgreSQLParser.g4
```diff
@@ -5401,7 +5401,7 @@ plsql_unreserved_keyword
    //| COLUMN_NAME
    | COMMIT
    | CONSTANT
-   | CONSTRAINT
```
huydo862003, line 5404: Kindly refer to this: https://www.notion.so/holistics/TDD-72-Database-CHECK-constraint-28bf89dc7e49801fa8d3fb9cf25ca8de?source=copy_link#291f89dc7e498014807cdf82224f0637

huyphung1602: Comment the TDD-72 to the code is okay in this case

packages/dbml-core/src/parse/schemarb/parser.pegjs
```diff
@@ -22,6 +22,21 @@
     }
   }
 
+  function addCheckConstraintToTable(tableName, expression, name) {
+    const table = data.tables.find(t => t.name === tableName);
+    if (!table) {
+      error("Table " + tableName + " not found");
```
huyphung1602: Should use template literal here instead em:
```javascript
error(`Table ${tableName} not found`)
```

PR #766, opened 2025-10-20
Feat/dbml connector/support check constraints

huydo862003:
## Summary
* Continuation of #765.
* Support extracting CHECK constraints in @dbml/connector.
* Remove support of converting CHECK constraints into DBML ENUMs in MSSQL connector.
* IMPORTANT: The MySQL connector currently generates malformed DBML when there's a check constraint in the SQL, because MySQL always store check constraints  with quoted column names in backticks. For example, this:
  ```
   CHECK (`col` LIKE '%...%')
  ```
  would be transformed into
  ```
   constraint: ``col` LIKE '%...%'`
  ```
  which is erroneous. We have no proper way to workaround this as we currently do not support escaping inside function expressions.

## Issue
(issue link here)

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-connector/src/connectors/mssqlConnector.ts
+++ b/packages/dbml-connector/src/connectors/mssqlConnector.ts
@@ -1,4 +1,9 @@
 /* eslint-disable camelcase */
+/* TDD 72
+ * - In MSSQL, there's no notion of ENUM.
+ * - MSSQL connector used to extract ENUMS from CHECK constraints, for example: `CHECK ([col]=value1 OR [col]=value2)` would be translated to an enum with 2 values, `value1` and `value2`.
+ * - Since DBML 4.0.0, we have supported CHECK constraints, so we no longer extract ENUMS out from CHECK constraints but rather keep them as CHECK constraints.
+ * */
 import sql from 'mssql';
 import { buildSchemaQuery, parseConnectionString } from '../utils/parseSchema';
 import {
@@ -560,6 +565,7 @@ const fetchSchemaJson = async (connection: string): Promise<DatabaseSchema> => {
   return {
     tables,
     fields,
+    // MSSQL doesn't support the notion of enums
     enums: [],
     refs,
     indexes,


packages/dbml-cli/__test__/db2dbml/mysql/expect-out-files/schema.dbml
```diff
@@ -136,9 +136,13 @@ Table "orders" {
...
   "id" "int unsigned" [pk, not null, increment]
   "price" decimal(10,2) [not null]
-  "quantity" int [not null]
-  "total_value" "decimal(10,2) GENERATED ALWAYS AS ((`price` * `quantity`)) STORED"
+  "quantity" int [not null, constraint: ``quantity` < 10000`]
+  "total_value" "decimal(10,2) GENERATED ALWAYS AS ((`price` * `quantity`)) STORED" [constraint: ``total_value` > 0`]
   "updated_at" "timestamp on update CURRENT_TIMESTAMP" [default: `CURRENT_TIMESTAMP`]
+
+  Constraints {
+    `(`price` > 0) and (`price` < 1000000)` [name: 'products_chk_1']
```
H-DNA: Please note that we currently do not escape `` ` `` inside function expression. Mysql always quote columns in backticks, so this is currently malformed DBML.

H-DNA: https://github.com/holistics/dbml/pull/766/files#r2458890208

packages/dbml-cli/__test__/db2dbml/mysql/expect-out-files/schema.dbml
```diff
@@ -136,9 +136,13 @@ Table "orders" {
...
   "id" "int unsigned" [pk, not null, increment]
   "price" decimal(10,2) [not null]
-  "quantity" int [not null]
-  "total_value" "decimal(10,2) GENERATED ALWAYS AS ((`price` * `quantity`)) STORED"
+  "quantity" int [not null, check: ``quantity` < 10000`]
+  "total_value" "decimal(10,2) GENERATED ALWAYS AS ((`price` * `quantity`)) STORED" [check: ``total_value` > 0`]
   "updated_at" "timestamp on update CURRENT_TIMESTAMP" [default: `CURRENT_TIMESTAMP`]
+
+  Checks {
+    `(`price` > 0) and (`price` < 1000000)` [name: 'products_chk_1']
```
huydo862003, line 144: Please note that we currently do not escape ` inside function expression. Mysql always quote columns in backticks, so this is currently malformed DBML.

https://www.notion.so/holistics/TDD-72-Database-CHECK-constraint-in-DBML-28bf89dc7e49801fa8d3fb9cf25ca8de?source=copy_link#292f89dc7e498001bbfce3d92f74cdd4

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyphung1602, line 1: Should add this related doc into this connector code as a comment em. You can also write some short notes for better understanding: https://www.notion.so/holistics/TDD-72-Database-CHECK-constraint-in-DBML-28bf89dc7e49801fa8d3fb9cf25ca8de#292f89dc7e498001bbfce3d92f74cdd4

packages/dbml-cli/__test__/cli.test.js
```diff
@@ -52,6 +52,6 @@ describe('@dbml/cli', () => {
 
   test.each(scanDirNames(__dirname, 'db2dbml'))('db2dbml/%s', async (dirName) => {
     await runTest(path.join(__dirname, 'db2dbml', dirName), '../bin/db2dbml.js');
-  }, 10000);
+  }, 100000);
```
huyphung1602: Can we user `const` to explain what is this number? Moreover, why do we need to increase this one x10 times 🤔

huydo862003: Yeah okay anh, it's the timeout for tests. For some reason, the test timed out so I needed to increase the timeout.

packages/dbml-connector/src/connectors/postgresConnector.ts
```diff
@@ -421,32 +422,92 @@ const generateIndexes = async (client: Client, schemas: string[]) => {
...
+      column_name,
+      is_column_constraint,
+    } = row;
+    const key = `${table_schema}.${table_name}`;
+    if (is_column_constraint) {
+      if (!tableConstraints[key]) tableConstraints[key] = {};
+      if (!tableConstraints[key][column_name]) tableConstraints[key][column_name] = { checks: [] };
+      tableConstraints[key][column_name].checks.push({
+        name: check_name || undefined,
+        expression: check_definition.slice(8, -2),
```
huyphung1602, line 507: What is 8 and what is -2, why do we choose this number 🤔 ? I suggest to use const, or you can also re-write a function just to cut this string. It will help the maintainer to learn the code easier.

packages/dbml-connector/src/connectors/postgresConnector.ts
```diff
@@ -421,32 +422,92 @@ const generateIndexes = async (client: Client, schemas: string[]) => {
...
+      });
+      return;
+    }
+    if (!checks[key]) checks[key] = [];
+    checks[key].push({
+      name: check_name || undefined,
+      expression: check_definition.slice(8, -2),
```
huyphung1602, line 516: Same here

packages/dbml-connector/src/connectors/mssqlConnector.ts
```diff
@@ -591,61 +468,99 @@ const generateIndexes = async (client: sql.ConnectionPool, schemas: string[]) =>
...
+      table_name,
+      column_name,
+      check_name,
+      check_expression,
+    } = row;
+    const key = `${table_schema}.${table_name}`;
+    if (column_name) {
+      if (!tableConstraints[key]) tableConstraints[key] = {};
+      if (!tableConstraints[key][column_name]) tableConstraints[key][column_name] = { checks: [] };
+      tableConstraints[key][column_name].checks.push({ name: check_name, expression: check_expression.slice(1, -1) });
```
huyphung1602, line 531: Same as the `.slice(8,-2)` comment

packages/dbml-connector/src/connectors/postgresConnector.ts
huydo862003: hmm maybe i will add a comment nha anh :pepe-cuoi-guong:. I don't think a variable name alone can explain this.

packages/dbml-connector/src/connectors/mssqlConnector.ts
huyphung1602: Hmm, but `slice(1, -1)` is more obvious, it may be acceptable 🤔

packages/dbml-connector/src/connectors/postgresConnector.ts
huyphung1602: If I understand this one correctly:

```
// PostgreSQL (line 496, 503)
expression: check_definition.slice(8, -2)
// Assumes format: "CHECK ((expression))" - exactly 8 chars prefix, 2 chars suffix

// MSSQL (line 524, 528)
expression: check_expression.slice(1, -1)
// Assumes format: "(expression)" - exactly 1 char on each side

// MySQL (line 428, 435)
expression: checkClause.slice(1, -1)
// Assumes format: "(expression)" - exactly 1 char on each side
```

huyphung1602: If yes, I believe you should use pattern matchicg/regex (not sure which technique, to validate if the string match the expression format before slicing. If not, it is buggy em.

My suggestion: Add null checks and validate length before slicing 🤔

huydo862003: I have added some comments nha

packages/dbml-connector/src/connectors/mysqlConnector.ts
```diff
@@ -386,7 +388,54 @@ async function generatePrimaryAndUniqueConstraint (client: Connection, schemaNam
     return acc;
   }, {});
 
-  return { compositeConstraintMap, constraintMap };
+  return { compositeConstraintMap, primaryAndUniqueConstraintMap: constraintMap };
+}
+
+function extractColumnsFromCheckConstraint (checkClause: string): string[] {
+  return [...checkClause.matchAll(/`(?:``|[^`])*`/g)].map((col) => col.toString().slice(1, -1));
```
huyphung1602, line 395: Crashes if `checkClause` is null. No validation of match results before slicing.

packages/dbml-connector/src/connectors/mysqlConnector.ts
```diff
@@ -386,7 +388,54 @@ async function generatePrimaryAndUniqueConstraint (client: Connection, schemaNam
...
+  const checks: CheckConstraintDictionary = {};
+  const tableConstraints: TableConstraintsDictionary = {};
+  rows.forEach((row: any) => {
+    const {
+      constraintName,
+      tableName,
+      checkClause,
+    } = row;
+    const key = tableName;
+    const cols = extractColumnsFromCheckConstraint(checkClause);
```
huyphung1602, line 421: Hmm I am assuming that the current code here only finds backtick-quoted identifiers. So `CHECK (age > 18)` gets treated as table-level because it has no backticks, even though it's a single column constraint.

**Fix:** Either:
1. Treat all as table-level (safe but less precise)
2. Query `information_schema.constraint_column_usage` to get the actual column associations

packages/dbml-connector/src/connectors/postgresConnector.ts
```diff
@@ -421,32 +422,92 @@ const generateIndexes = async (client: Client, schemas: string[]) => {
...
+  const checks: CheckConstraintDictionary = {};
+  checkListResult.rows.forEach((row) => {
+    const {
+      table_schema,
+      table_name,
+      check_name,
+      check_definition,
+      column_name,
+      is_column_constraint,
+    } = row;
```
huyphung1602, line 489: Not sure if we have validation after de-structure. If any field is undefined, we may create bad data silently (I believe our legacy code here has a bad practice when de-structure object too).

packages/dbml-connector/src/connectors/mysqlConnector.ts
huydo862003: - Isn't `checkClause` already annotated with `string` anh?
- Further note: `matchAll` will always return an iterator, even if there's no match, in which case, an empty iterator will be return. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll.

packages/dbml-connector/src/connectors/mysqlConnector.ts
huydo862003: - MySQL will normalize the check constraint clause before storing in the database, so the check clause always contains columns quoted in backticks anh. [Claude](https://claude.ai/share/63a707ba-0741-48c8-ba30-306b4bc97ffa). -> Our solution is safe.
- The reason I did this was because in MySQL, there's no table named `information_schema.constraint_column_usage` or any other table storing column constraints. I assume MySQL converts them all to table constraints. -> Solution 2 wouldn't work.

packages/dbml-connector/src/connectors/postgresConnector.ts
huydo862003: Added validation nha anh

packages/dbml-connector/src/connectors/postgresConnector.ts
huydo862003: Added validation nha anh

packages/dbml-connector/src/connectors/mssqlConnector.ts
huydo862003: Done!

Issue #761, opened 2025-09-30
Unable to provide default value as enum value reference


nazmul-bzs:
1. I was trying to provide a default value of the status field of appointment table to the enum reference AppointmentStatusType.Requested. But I got compiler error. It was okay if I provide just the string literal "Requested". It would be better if it allow us to access enum reference directly.

<img width="896" height="462" alt="Image" src="https://github.com/user-attachments/assets/44142fde-3fee-45dc-a15a-9d70b65d1268" />

<img width="907" height="394" alt="Image" src="https://github.com/user-attachments/assets/37221c0e-961d-4603-90d1-bc33dabe7f5a" />

2. Also, there is no compiler error if we provide any default value that does not belong to the enum.

<img width="892" height="397" alt="Image" src="https://github.com/user-attachments/assets/a15ebc69-8469-4e09-b988-a2ee599df2f4" />

huyleminh01:
Hi @nazmul-bzs ,

Thank you for reporting these enum-related issues.

For me, these improvement about enum referencing can help us reduce the effort to maintain the enum value in the whole DBML project.

Let me take note of your request so the team can consider improving them in the future.

Also, if possible, you can contribute these fixes by opening the PR, and we will review for you.

Thanks.

huydo862003:
Dear @nazmul-bzs, regarding:
1. Providing an enum value into a column's default is now **supported**:
    - On dbdiagram, you can now use this feature.
    - If you're using npm package from the dbml repository, make sure you upgrade to the latest version (v5.2.0).
2. Validating whether a value belongs to an enum:
    - I believe that this use case primarily stems from your inability to provide an enum value into a column's default.
    - For other data types (integers, texts), we also currently do not provide any typechecking.
    Therefore, this is currently not prioritized.

Should you have any questions, feel free to drop a comment or a new issue!
Thank you

PR #777, opened 2025-11-06
Feat/support enum as default value

huydo862003:
## Summary
* Support enum as default value in column settings.
  * No changes to @dbml/core and model_structure
  * When an enum is provided as a default value, the interpreter will transform the enum value into a string and yield a default value of type string in model_structure. For example:
    ```js
       Enum schema.enum {
           value1
       }

       Table table {
            col schema.enum [default: schema.enum.value1] // translated into { dbdefault: { value: 'value1', type: 'string' }}
       }
    ```
* DBX-6346

## Issue
#761

## Checklist

- [ ] Code Review

### DIFF
--- a/packages/dbml-parse/src/lib/analyzer/binder/elementBinder/elementBinder.ts
+++ b/packages/dbml-parse/src/lib/analyzer/binder/elementBinder/elementBinder.ts
@@ -257,7 +257,7 @@ export default abstract class ElementBinder {
       this.logError(
         accessSubname.referrer,
         `Can not find ${kind} '${name}'`,
-        rule.ignoreNameNotFound || (rule.ignoreNameNotFoundForQuotedVariable && isFirstFragmentQuotedVariable),
+        rule.ignoreNameNotFound || (!!rule.ignoreNameNotFoundForQuotedVariable && isFirstFragmentQuotedVariable),
       );
       return;
     }


packages/dbml-parse/src/lib/analyzer/binder/types.ts
```diff
@@ -23,6 +23,9 @@ export type BinderRule =
     ignoreNameNotFound: boolean;
     // A list of keywords that, when encountered by the binder as a standalone variable, skipped
     keywords?: readonly string[];
+    // In attribute value, a quoted variables can be sometimes used as a string
+    // This option decides whether it is legal for a standalone quoted variable to be treated as a string and not bound
+    ignoreNameNotFoundForQuotedVariable: boolean;
```
huyphung1602: In order to prevent numerous declarations in other code, should we set this one's default value to false?

huydo862003: Should I declare it as optional, that way it's undefined (false) by default?
```
    ignoreNameNotFoundForQuotedVariable?: boolean;
```
However, because I have defined ignoreNameNotFoundForQuotedVariable in the other code, I think would keep them instead of removing.

huydo862003: I have made it optional nhe

packages/dbml-parse/src/lib/analyzer/binder/elementBinder/elementBinder.ts
```diff
@@ -211,11 +211,13 @@ export default abstract class ElementBinder {
         : extractVarNameFromPrimaryVariable(fragment).unwrap();
 
       subnameStack.unshift({
-        index: createNodeSymbolIndex(varname, remainingSubnamesSymbolKind!),
+        index: createNodeSymbolIndex(varname, rule.remainingSubnamesSymbolKind!),
         referrer: fragment,
       });
     }
 
+    const isFirstFragmentQuotedVariable = isExpressionAQuotedString(rawFragments[0]);
```
huyphung1602, line 219: Nice addition of `isFirstFragmentQuotedVariable` here—it'll help skip binding errors for quoted strings in attributes. Does this cover all cases, like nested quotes?

huydo862003: By nested quotes, do you mean like `'abc\'def'` :thinking: If so, yeah, it would handle this case.

packages/dbml-parse/src/lib/analyzer/binder/elementBinder/elementBinder.ts
```diff
@@ -228,16 +230,22 @@ export default abstract class ElementBinder {
...
+      : this.resolveIndexStack(
+          subnameStack,
+          rule,
+          isFirstFragmentQuotedVariable,
+        );
   }
 
   // Looking up the indexes in the subname stack from the current declaration node
   // Each time the index resolves to a symbol, the referrer's symbol is bound to it
   private resolveIndexStack (
```
huyphung1602, line 245: The resolveIndexStack method now has 3 params, increasing complexity. Without better docs, it's hard to follow—add a JSDoc comment explaining rule and `isFirstFragmentQuotedVariable` to prevent future confusion or bugs.

PR #800, opened 2026-01-16
Support sample table data in DBML

huydo862003:
## Summary

Add support for sample table data in DBML.

Doc: https://github.com/holistics/dbml/blob/78d79441758adc973376a18531b8d0b3807f9208/dbml-homepage/docs/docs.md
TDD: https://www.notion.so/holistics/TDD-80-Sample-Table-Data-2e6f89dc7e498073ac17c09a842aa64b

### Support CommaExpressionNode in parser
- Added `CommaExpressionNode` to AST for comma-separated value lists
- Added lexer support for scientific notation (e.g., `1.23e5`, `-4.56e-3`)

### Support sample table data in DBML

```dbml
Table Users {
  id string [pk]
  name string
  age int

  Records { // implicit columns
    "id1", "Alice", 18
    "id2", "Bob", 19
  }

  Records (id, name) { // explicit columns
    "id3", "Charlie"
  }
}

Records Users(id, name) { // always explicit
  "id4", "David"
}
```

- Implicit/explicit column lists
- Full type validation and constraint checking (PK, FK, unique, not null)
- Support for strings, numbers, booleans, dates, enums, null, and SQL expressions
- Validation warnings instead of errors for constraint violations

### Handle INSERT statements in importers and exporters

**Exporters:** Convert `Records` to `INSERT` statements for all SQL dialects (MySQL, PostgreSQL, Oracle, MSSQL, Snowflake) with automatic FK constraint handling

**Importers:** Parse `INSERT` statements from SQL and convert to DBML `Records`, including Snowflake support

### Add diagnostics provider for reusable logic related to errors and warnings

- New `DBMLDiagnosticsProvider` class with `provideDiagnostics()`, `provideErrors()`, `provideWarnings()` methods

### Others

- Hardcode Report<T, E> to Report<T>
  - `E` is now always `CompileError`, also `CompileWarning` is supported
  - Simplified API with `getErrors()` and `getWarnings()` methods

- Expose reusable SQL type checking and value extraction functions
  - Type checkers: `isSqlNumericType()`, `isSqlStringType()`, `isSqlDateTimeType()`, `isSqlBooleanType()`
  - Value extractors: `tryExtractBoolean()`, `tryExtractInteger()`, `tryExtractFloat()`, `tryExtractString()`, `tryExtractDate()`, `tryExtractDateTime()`, `tryExtractTime()`
  - Utilities: `normalizeRecordValue()`, `addDoubleQuoteIfNeeded()`

## Issue

(issue link here)

### DIFF
--- a/packages/dbml-core/src/export/OracleExporter.js
+++ b/packages/dbml-core/src/export/OracleExporter.js
@@ -5,6 +5,8 @@ import {
   buildUniqueTableName,
   escapeObjectName,
   shouldPrintSchema,
+  parseIsoDatetime,
+  formatDatetimeForOracle,
 } from './utils';
 import {
   isNumericType,
@@ -38,7 +40,26 @@ class OracleExporter {
 
         if (isNumericType(val.type)) return val.value;
         if (isBooleanType(val.type)) return String(val.value).toUpperCase() === 'TRUE' ? '1' : '0';
-        if (isStringType(val.type) || isDateTimeType(val.type)) return `'${val.value.replace(/'/g, "''")}'`;
+
+        // Handle datetime types with TO_TIMESTAMP
+        if (isDateTimeType(val.type)) {
+          const datetimeInfo = parseIsoDatetime(val.value);
+          if (datetimeInfo) {
+            const { datetime, hasTimezone } = datetimeInfo;
+            const formattedDatetime = formatDatetimeForOracle(datetime, hasTimezone);
+
+            if (hasTimezone) {
+              // Use TO_TIMESTAMP_TZ for timestamps with timezone
+              return `TO_TIMESTAMP_TZ('${formattedDatetime}', 'YYYY-MM-DD HH24:MI:SS.FF3 TZH:TZM')`;
+            }
+            // Use TO_TIMESTAMP for timestamps without timezone
+            return `TO_TIMESTAMP('${formattedDatetime}', 'YYYY-MM-DD HH24:MI:SS.FF3')`;
+          }
+          // Fallback to quoted string if parsing fails
+          return `'${val.value.replace(/'/g, "''")}'`;
+        }
+
+        if (isStringType(val.type)) return `'${val.value.replace(/'/g, "''")}'`;
         if (isBinaryType(val.type)) return `HEXTORAW('${val.value}')`;
         // Unknown type - use CAST
         return `CAST('${val.value.replace(/'/g, "''")}' AS ${val.type})`;
--- a/packages/dbml-core/__tests__/examples/exporter/oracle_exporter/output/insert_records.out.sql
+++ b/packages/dbml-core/__tests__/examples/exporter/oracle_exporter/output/insert_records.out.sql
@@ -19,9 +19,9 @@ ALTER TABLE "posts" ADD FOREIGN KEY ("user_id") REFERENCES "users" ("id") DEFERR
 SET CONSTRAINTS ALL DEFERRED;
 
 INSERT ALL
-  INTO "users" ("id", "name", "email", "active", "created_at") VALUES (1, 'Alice', 'alice@example.com', 1, '2024-01-15T10:30:00+07:00')
-  INTO "users" ("id", "name", "email", "active", "created_at") VALUES (2, 'Bob', 'bob@example.com', 0, '2024-01-16T14:20:00+07:00')
-  INTO "users" ("id", "name", "email", "active", "created_at") VALUES (3, 'Charlie', NULL, 1, '2024-01-17T09:15:00+07:00')
+  INTO "users" ("id", "name", "email", "active", "created_at") VALUES (1, 'Alice', 'alice@example.com', 1, TO_TIMESTAMP_TZ('2024-01-15 10:30:00.000 +07:00', 'YYYY-MM-DD HH24:MI:SS.FF3 TZH:TZM'))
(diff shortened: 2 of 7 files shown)


packages/dbml-parse/src/core/interpreter/interpreter.ts
```diff
@@ -31,7 +70,7 @@ export default class Interpreter {
   ast: ProgramNode;
   env: InterpreterDatabase;
 
-  constructor (ast: ProgramNode) {
+  constructor (ast: ProgramNode, source: string) {
```
NQPhuc: Seeing that more and more places in the parser require the source for context, should the source be placed inside ast? (I'm not suggesting we should do it right now, just want to leave a note)

huydo862003: Updated in this nhe:

https://github.com/holistics/dbml/pull/800/changes/a9480b081d2d261ac72813732fabfe625e3fe08b

packages/dbml-parse/src/core/parser/nodes.ts
```diff
@@ -413,6 +466,41 @@ export class TupleExpressionNode extends SyntaxNode {
...
+// e.g. 1, 2, (trailing comma)
+export class CommaExpressionNode extends SyntaxNode {
+  elementList: NormalExpressionNode[];
+
+  commaList: SyntaxToken[];
+
+  constructor (
+    {
+      elementList = [],
+      commaList = [],
```
NQPhuc, line 485: Does the elementList and commaList always have the same lenght? If the expression is `a,,b,c` what will they look like?

NQPhuc: Also, does the commaList have any sematic meaning or do we need it for any thing?

huydo862003: elementList and commaList will have interleaved elements.

```
a,,b,c
```
Can be roughly translated to:
```
<1st element (a)><first comma><2nd element (empty node)><second-comma>...
```

We need commaList for the syntax tree to be lossless, which is needed for IDE-centric parser - We need to preserve all tokens in the tree so we can locate where we're currently at while editing the code for example.

NQPhuc: Ah so they're used by the IDE services. We can keep them for now

packages/dbml-core/__tests__/examples/importer/oracle_importer/input/records.in.sql
```diff
@@ -0,0 +1,37 @@
...
+
+-- Multiple INSERT statements for employees table
+INSERT INTO "employees" ("id", "dept_id", "name")
+VALUES (1, 1, 'John Doe');
+
+INSERT INTO "employees" ("id", "dept_id", "name")
+VALUES (2, 1, 'Jane Smith');
+
+INSERT INTO "employees" ("id", "dept_id", "name")
+VALUES (3, 2, 'Bob Johnson');
```
NQPhuc, line 37: Look like when exporting dbml->oracle sql, you export as `INSERT ALL...`. But when import oracle sql -> dbml, you only tested for `INSERT INTO` statements. Please help re-check and add test for `INSERT ALL`

packages/dbml-core/src/export/OracleExporter.js
```diff
@@ -6,8 +6,62 @@ import {
...
+          const valueStrs = row.map(formatValue);
+          return `  INTO ${tableRef} ${columnList} VALUES (${valueStrs.join(', ')})`;
+        });
+        return `INSERT ALL\n${intoStatements.join('\n')}\nSELECT * FROM dual;`;
+      }
+
+      // Single row INSERT
+      const valueStrs = values[0].map(formatValue);
+
+      return `INSERT INTO ${tableRef} ${columnList}\nVALUES (${valueStrs.join(', ')});`;
```
NQPhuc, line 80: Please make sure the Oracle importer can processe `INSERT ALL` to maintain exporter/importer consistence. Alternatively, you could also use INSERT INTO for all cases like the other exporter

packages/dbml-core/src/export/DbmlExporter.js
```diff
@@ -347,6 +347,45 @@ class DbmlExporter {
...
+        const allColumnIndexes = allColumns.map((col) => record.columns.indexOf(col));
+        return record.values.map((row) => allColumnIndexes.map((colIdx) => colIdx === -1 ? { value: null, type: 'expression' } : row[colIdx]));
+      });
+
+      // Build data rows
+      const rowStrs = allRows.map((row) =>
+        `  ${row.map(formatRecordValue).join(', ')}`,
+      );
+
+      return `records ${tableRef}(${columnList}) {\n${rowStrs.join('\n')}\n}\n`;
```
NQPhuc:
```suggestion
      return `Records ${tableRef}(${columnList}) {\n${rowStrs.join('\n')}\n}\n`;
```

huydo862003: Fixed in https://github.com/holistics/dbml/pull/800/changes/e443e7d229fbd0016b512b14175401fe9290321a nha anh

packages/dbml-core/src/export/OracleExporter.js
huydo862003: fixed nhe anh: https://github.com/holistics/dbml/pull/800/changes/2ef8f52259ba889fde3156c589c02c6358b687ff

Supported INSERT ALL

Issue #810, opened 2026-01-30
Bug : parsing issue when exporting a Postgresql GENERATED field


kbsali:
When trying to export a Postgresql table dump to dbml I get the following error :
```
$ sql2dbml docs/db/01-tables.generated.sql --postgres 
  ERROR: 
    You have a syntax error at "01-tables.generated.sql" line 394 column 41. mismatched input '(' expecting ')'
```
And this is the problematic line : 
```
...
email text GENERATED ALWAYS AS (lower((identity_data ->> 'email'::text))) STORED,
...
```
The "temporary" fix to make it work :
```
...
email text,
...
```

huyphung1602:
Hi there,

Thank you for providing such a detailed report.

We are investigating this issue and will notify you as we get an update.

NQPhuc:
Hi @kbsali.

Thank you for the report.

We have released a new version of dbml (6.0.0), which should resolve this issue. Feel free to try it out and let us know how it turns out.

kbsali:
just tested, it works great!👍 
Thanks!!!

Issue #553, opened 2024-04-16
importing posgresql function make import fail


pierresouchay:
When parsing any function with BEGIN/END, parsing will fail:

Small example:
```sql
CREATE FUNCTION public.my_func_max_42() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
        BEGIN
          IF (SELECT SUM(mycounter::int) FROM mytable) > NEW.mycounter::int > 42
          THEN Raise Exception 'Sum must be lower than 42';
          END IF;
          RETURN NEW;
        END;
        $$;
```

=> This make the whole process fail, likely due to `;` chars within the function

pierresouchay:
For now, my only workaround is to remove lots of stuffs from the SQL schema using the following script:

```python
#!/usr/bin/env python3

import os
import re
import sys


"""
An ugly script to pre-process sql and to have an output suitable for sql2dbml
"""

FUNCTION_PATTERN = re.compile("^(CREATE|ALTER) FUNCTION")
VIEW_PATTERN = re.compile("^CREATE.* VIEW")
AS_PATTERN = re.compile("[\\s]*AS ([^\\s]+)")
JSON_ACCESSOR = re.compile(" ->> '([^']+)'")

searching_for = None
current_function = ""


def print_ignored_line(val: str) -> None:
    if "DEBUG" in os.environ:
        print(f"[IGNORED] {val}", file=sys.stderr)


with open(sys.argv[1], "r") as f:
    for line in f.readlines():
        if searching_for:
            current_function += line
            if line.strip().endswith(searching_for):
                searching_for = None
                print_ignored_line(current_function)
                current_function = ""
            elif AS_PATTERN.match(line):
                symbol = AS_PATTERN.match(line).group(1)
                searching_for = f"{symbol};"
        elif FUNCTION_PATTERN.match(line) or VIEW_PATTERN.match(line):
            current_function = line
            if not line.strip().endswith(";"):
                searching_for = ";"
            else:
                print_ignored_line(line)
        else:
            # sql2dbml does not like postgresql JSON Accessors
            # replace ->> with .
            print(JSON_ACCESSOR.sub(".\\1", line), end="")
```

simonschmidt:
I hit the same issue, since I produce sql with `pg_dump` to begin with I was able to use  `--disable-dollar-quoting`

```
pg_dump \
  --schema-only \
  --disable-dollar-quoting \
  "$DB_URL" |
  sql2dbml --postgres /dev/stdin
```

[...]

Issue #436, opened 2023-09-07
Postgresql parsing fail on function body: no viable alternative at input


mermetbt:
Hello,

Thanks for this PR https://github.com/holistics/dbml/pull/416

I've noticed the parser is not working on function like this:

```sql
CREATE FUNCTION public.myfunction() RETURNS void
    LANGUAGE sql
    AS $$
        -- Do something
    $$;
```

with the output being:

```
Error: no viable alternative at input '$$\n\t-- Do something\n            $$;\n'
```

Regards.

mermetbt:
As a workaround, I use this simple sed command to remove the functions from the sql dump `sed '/CREATE FUNCTION/,/\$\$;/d' input.sql > output.sql`

NQPhuc:
Hi mermetbt, 

Thanks for the workaround. Indeed the new postgres grammar introduce in https://github.com/holistics/dbml/pull/416 even though more complete than our previous parser, it's still not perfect. We'll try to improve this in the future.

NQPhuc:
Fixed in https://github.com/holistics/dbml/pull/812

PR #812, opened 2026-02-03
Fix: update postgres lexer and base class to support json operators

NQPhuc:
## Summary
* Update postgres lexer and parser g4 and base classes to latest to support multi ops operators (->>, #>>, #>,...)
* Move the postgres base classes to `parsers` folder because they are supposed to be next to the generated lexer and parser files

## Issue
* https://github.com/holistics/dbml/issues/810
* The update parser grammar also resolve these: 
  * https://github.com/holistics/dbml/issues/436
  * https://github.com/holistics/dbml/issues/553

## Checklist

- [ ] Lint Checks Passed
- [x] Unit Tests Passed
- [ ] Coverage Tests Passed
- [ ] Code Review

### DIFF
--- a/packages/dbml-core/__tests__/examples/importer/postgres_importer/input/create_function.in.sql
+++ b/packages/dbml-core/__tests__/examples/importer/postgres_importer/input/create_function.in.sql
@@ -1,0 +1,22 @@
+-- https://github.com/holistics/dbml/issues/436
+CREATE FUNCTION public.myfunction() RETURNS void
+    LANGUAGE sql
+    AS $$
+        -- Do something
+    $$;
+-- https://github.com/holistics/dbml/issues/553
+CREATE FUNCTION public.my_func_max_42() RETURNS trigger
+    LANGUAGE plpgsql
+    AS $$
+        BEGIN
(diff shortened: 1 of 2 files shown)


packages/dbml-core/src/parse/ANTLR/parsers/postgresql/PostgreSQLParserBase.js
huydo862003, line 1: Since you updated the grammar, in this case, this file should also be updated? 👀 

https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/JavaScript/PostgreSQLParserBase.js

NQPhuc: I only update the lexer.g4 and lexer base clase and didn't touch parser.g4 or parser base class since changing them would require a bigger refactor

NQPhuc: I updated the parser in this PR: https://github.com/holistics/dbml/pull/814

PR #870, opened 2026-04-14
DBX-6700 Query-based compiler & Multifile support

huydo862003:
## Summary
* BREAKING CHANGE: Revamp the compiler into query-based design.
* Add multifile support into compiler:
  ```dbml
  use { table my_table as t } from './a.dbml'
  use * from './b.dbml'
  ```
  * `use`: Non-transitive imports
  * `reuse`: Transitive imports
  * Allow importing of: `Table`, `Enum`, `TableGroup`, `Schema`, `Note` (sticky note), `TablePartial`
    * `Ref`, `Records` are automatically pulled if all of its tables are in scope.
    * `TableGroup` import would also import all of its tables. `TableGroup` cannot reference imported table, it can only reference the current file's table (for now).
    * `Schema` import would pull all elements inside that schema into scope.

* Discussion: https://github.com/holistics/dbml/discussions/843
* TDD:
  - Query-based compiler: https://www.notion.so/holistics/TDD-94-Query-based-compiler-Overview-of-the-compiler-architecture-345f89dc7e4980e88bc9fd6d140c1d8b
  - Multifile support (reworked): https://www.notion.so/holistics/TDD-91-DBML-Multi-file-Support-319f89dc7e49804f94f6d8d7cd2cdf6e?v=b4c7e791a9664e779faf139f7f7bda40&source=copy_link
  
## Suggested review order
1. Review the TDD for query-based compiler: https://www.notion.so/holistics/TDD-94-Query-based-compiler-Overview-of-the-compiler-architecture-345f89dc7e4980e88bc9fd6d140c1d8b
2. Check the files for `Filepath`, `SyntaxNode`, `NodeSymbol` and the internable types `Internable`.
3. Check the `compiler.ts` file:
   a. Check how the local cache & the global cache are setup (`localQuery` and `globalQuery` methods).
   b. Check what queries we have: https://www.notion.so/holistics/TDD-94-Query-based-compiler-Overview-of-the-compiler-architecture-345f89dc7e4980e88bc9fd6d140c1d8b?v=b4c7e791a9664e779faf139f7f7bda40&source=copy_link#345f89dc7e49804f8f0edca6282a32c8
4. Check the overall flow of the compiler: Compiler query -> Interning the args -> Check query cache -> Dispatch to query module (chain-of-responsibility) -> Cache the result -> return.
4. Check `lexer`, `parser`.
5. Check the `local_modules` for validator (do not care about multifile support yet).
[...]

PR #883, opened 2026-04-27
Feat/playground rewrite

huydo862003:
## Summary

- Rewrite `dbml-playground`.
- Move some changes from #870 to reduce changed files on that branch


dbml-playground/src/components/SplitPanel.vue
```diff
@@ -0,0 +1,86 @@
...
+
+<script setup lang="ts">
+import {
+  ref, computed,
+} from 'vue';
+
+// This component creates a resizable set of panes
+
+const {
+  minSize = 40, // The minimum size in fr allowed for a pane
```
xuantho573: This value will make the right-most pane being too small when resizing the other panes, consider reducing this value

huydo862003: fix nhóa

xuantho573 (approved):

PR #912, opened 2026-05-28
Hotfix: Infinite loop when running use-then-create-table mssql in dbml

huydo862003:
## Summary
- Problem: There's an infinite loop in the parser triggered by an MSSQL syntax appearing inside DBML (e.g `USE [db] GO CREATE TABLE [schema].[table](...)`)
- Root cause: The `synchronizeUseSpecifier` error recovery loop could spin indefinitely when tokens didn't match its exit conditions (`}` or start-of-line). This is because it isn't ensured to always consume AT LEAST 1 token. So the loop can happen like: Parse -> Error -> Recover (does not consume any token) -> Parse the invalid token again -> ...
- Solution: Remove the `synchronizeUseSpecifier`, just throw the error to `useSpecifierList` for it to handle

## Issue
(issue link here)

## Checklist

- [ ] Lint Checks Passed
- [ ] Unit Tests Passed
- [ ] Coverage Tests Passed
- [ ] Code Review

### DIFF
--- a/packages/dbml-parse/src/core/parser/parser.ts
+++ b/packages/dbml-parse/src/core/parser/parser.ts
@@ -187,8 +187,10 @@ export default class Parser {
     this.tokens = tokens;
   }
 
-  parse (): Report<{ ast: ProgramNode;
-    tokens: SyntaxToken[]; }> {
+  parse (): Report<{
+    ast: ProgramNode;
+    tokens: SyntaxToken[];
+  }> {
     const body = this.program();
     const eof = this.advance();
     const program = this.nodeFactory.create(ProgramNode, {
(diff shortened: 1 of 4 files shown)


packages/dbml-parse/src/core/parser/parser.ts
```diff
@@ -418,10 +417,7 @@ export default class Parser {
         if (e.partialNode instanceof SyntaxNode) {
           args.name = e.partialNode;
         }
-        if (!this.canHandle(e)) {
-          throw new PartialParsingError(e.token, buildNode(), e.handlerContext);
-        }
-        this.synchronizeUseSpecifier();
+        throw new PartialParsingError(e.token, buildNode(), e.handlerContext); // Let use specifier list handle this
```
huyphung1602, line 434: Add some comments for Tho and AI?

packages/dbml-parse/src/core/parser/parser.ts
huyphung1602, line 403: should add comments to describe this function?

huyphung1602 (approved):

PR #923, opened 2026-06-04
DBX-6750 Support example records

huydo862003:
## Summary
* Support `example` flag on `records`
  ```
  records table_name [example] {}
  ``` 
* When a records have the example flag, it won't be exported into SQL.

## Issue
(issue link here)

## Checklist

- [ ] Lint Checks Passed
- [ ] Unit Tests Passed
- [ ] Coverage Tests Passed
- [ ] Code Review


packages/dbml-parse/src/core/local_modules/records/validate.ts
```diff
@@ -1,8 +1,9 @@
 import { partition } from 'lodash-es';
+import { forIn } from 'lodash-es';
```
xuantho573: merge with the import above nhe

packages/dbml-parse/src/core/global_modules/records/interpret.ts
```diff
@@ -91,11 +91,13 @@ export default class RecordsInterpreter {
     }
 
     const token = getTokenPosition(this.element);
+    const example = this.metadata.example(this.compiler) || undefined;
```
xuantho573, line 94: The method already return a boolean value, so I don't think we need an `undefined` here 🤔

huydo862003: the point is to avoid breaking tests :rofl: adding `example: false` would break existing tests for the snapshots

packages/dbml-parse/src/core/local_modules/records/validate.ts
huydo862003: [dcacb07](https://github.com/holistics/dbml/pull/923/commits/dcacb072459246db55058257df9471a6c705b3ed) fixed nhó

xuantho573 (approved):

PR #925, opened 2026-06-10
Add benchmark ci + optimize the validate records fk flow

huydo862003:
## Summary
* Add benchmark ci.
* Add script to run benchmark and profile for parser.
* Optimize the validate records fk flow, which take 50% of the time doing nothing on large diagrams

## Issue
(issue link here)

## Checklist

- [ ] Lint Checks Passed
- [ ] Unit Tests Passed
- [ ] Coverage Tests Passed
- [ ] Code Review


packages/dbml-parse/src/core/types/report.ts
```diff
@@ -31,8 +31,8 @@ export default class Report<T> {
```
xuantho573, line 31: Since the signature of the functions changed, please remember to update the comments as well

packages/dbml-parse/src/compiler/index.ts
```diff
@@ -190,7 +211,11 @@ export default class Compiler {
       this.cleanStaleLocalCache(filepath);
 
       const filepathId = filepath.intern();
-      const argKey = args.map((a) => intern(a)).join('\0');
+      const argKey = args.length === 0
+        ? ''
+        : args.length === 1
+          ? intern(args[0]) as string
+          : args.map((a) => intern(a)).join('\0');
```
xuantho573, line 218: Not sure how does this differ from the old code 🤔 as the output of both code should be the same, and the previous code looks cleaner to me 👀

packages/dbml-parse/src/compiler/index.ts
```diff
@@ -123,6 +130,8 @@ export default class Compiler {
   // (QuerySymbol, interned argument string) -> Result
   private globalCache = new Map<QuerySymbol, Map<string, any>>();
 
+  private globallyQuerying = false; // Check if we're already inside a query to skip unnecessary repeat global checks
```
xuantho573: What are global checks 🤔

packages/dbml-parse/src/core/types/report.ts
huydo862003: gke :O

https://github.com/holistics/dbml/pull/925/commits/0ce0b622c785aa69060232623ed4ec2654c07b61

packages/dbml-parse/src/compiler/index.ts
huydo862003: This code is run in the hot path (most compiler queries)

We avoid some list/string allocations to avoid GC blocks/memory pressure when args is 0 or 1 (most of the case)

Not sure if the engine would optimize this so not really sure :3

packages/dbml-parse/src/compiler/index.ts
huydo862003: https://github.com/holistics/dbml/pull/925/commits/e8f2f3cd366cdd989bc1c4ff1cb50144aa69b86f

xuantho573 (approved):
🙈

PR #933, opened 2026-06-22
Data dependency

NQPhuc:
## Summary

Adds the `Dep` element to the DBML language: a directed dependency between tables or columns.

```
Dep: users -> orders                        // table to table
Dep: orders.user_id <- users.id             // column to column, either direction
Dep: my_schema.events.id -> users.id        // schema-qualified endpoints
Dep: users -> reports [color: #abcdef]      // inline setting

Dep dep_header [color: #aabbcc] {           // named block, header setting
  users -> orders
}

Dep dep_body {                              // settings in the body, plus a note
  note: 'a comment body quotes the post title'
  orders -> events
  color: #123456
}
```

## Issue

https://github.com/holistics/dbml/discussions/879

## Checklist

- [ ] Lint Checks Passed
- [x] Unit Tests Passed
- [x] Coverage Tests Passed
- [ ] Code Review


packages/dbml-parse/src/core/local_modules/table/validate.ts
```diff
@@ -575,6 +586,13 @@ export function validateFieldSetting (parts: ExpressionNode[]): Report<Settings>
           }
         });
         break;
+      case SettingName.Dep:
+        attrs.forEach((attr) => {
+          if (!isUnaryRelationship(attr.value)) {
+            errors.push(new CompileError(CompileErrorCode.INVALID_COLUMN_SETTING_VALUE, '\'dep\' must be a valid unary relationship', attr.value || attr.name!));
+          }
+        });
+        break;
```
huydo862003: I suggest these two validations can use a reusable `isUnaryDependency` function like `isUnaryRelationship`.

Currnently, this doesn't raise any error, as you only validate the dep op, not the dep operands yet, for this, you can take `isUnaryRelationship` for reference. 👀 

Image

packages/dbml-parse/src/core/local_modules/dep/validate.ts
```diff
@@ -0,0 +1,127 @@
...
+      if (!sub.type) return [];
+      return this.compiler.validateNode(sub).getErrors();
+    });
+  }
+}
+
+export function validateDepSettings (settings: ListExpressionNode): Report<Settings> {
+  const aggReport = aggregateSettingList(settings);
+  return new Report(aggReport.getValue(), aggReport.getErrors());
+}
```
huydo862003, line 218: Does `dep` allow any setting list? if not, should disallow, see `CustomValidator.validateSettingList` for more details. 

Then this function should also be called inside `DepValidator.validate()` too, else the validate flow wouldn't catch it 👀

packages/dbml-parse/src/core/local_modules/custom/validate.ts
```diff
@@ -26,9 +26,9 @@ export default class CustomValidator {
   }
 
   private validateContext (): CompileError[] {
-    if (!(this.declarationNode.parent instanceof ElementDeclarationNode && this.declarationNode.parent.isKind(ElementKind.Project))) {
+    if (!(this.declarationNode.parent instanceof ElementDeclarationNode && this.declarationNode.parent.isKind(ElementKind.Project, ElementKind.Dep))) {
       return [
-        new CompileError(CompileErrorCode.INVALID_CUSTOM_CONTEXT, 'A Custom element can only appear in a Project', this.declarationNode),
+        new CompileError(CompileErrorCode.INVALID_CUSTOM_CONTEXT, 'A Custom element can only appear in a Project or a Dep', this.declarationNode),
```
huydo862003, line 31: Which custom element is allowed in Dep :V?

A custom element accept any name, such as:

```
Dep {
  a: "..." // A custom
  b: 3     // A custom
  
  Table.col -> Table.col // Not a custom
}
```

huydo862003:
Pls help run lint:fix nhe :v

packages/dbml-parse/src/core/global_modules/tablePartial/interpret.ts
```diff
@@ -221,6 +221,8 @@ export class TablePartialInterpreter {
       return inlineRef;
     });
 
+    column.inline_deps = [];
```
huydo862003, line 228: Can we have a comment like: columns in partial table do not support dep, so hardcode to `[]`

packages/dbml-parse/src/core/global_modules/table/index.ts
```diff
@@ -193,6 +193,16 @@ export const tableModule: GlobalModule = {
...
+    if (isInsideSettingValue(node, SettingName.Dep)) {
+      const enclosingDecl = node.parent;
+      if (enclosingDecl instanceof ElementDeclarationNode && enclosingDecl.isKind(ElementKind.Table)) {
+        const programNode = compiler.parseFile(node.filepath).getValue().ast;
+        const globalSymbol = compiler.nodeSymbol(programNode).getValue();
+        if (globalSymbol === UNHANDLED) return Report.create(undefined);
+        return nodeRefereeOfInlineDep(compiler, globalSymbol, node);
+      }
+    }
+
```
huydo862003: Why do we check for `isInsideSettingValue(Dep)` in 2 places? 👀 

According to the current pattern, we should add a check like below, not here?

packages/dbml-parse/src/core/global_modules/table/index.ts
```diff
@@ -221,6 +231,10 @@ export const tableModule: GlobalModule = {
       return nodeRefereeOfInlineRef(compiler, globalSymbol, node);
     }
 
+    if (isInsideSettingValue(node, SettingName.Dep)) {
+      return nodeRefereeOfInlineDep(compiler, globalSymbol, node);
+    }
+
     // Case 3: Column's default value being an enum value
```
huydo862003: Add a case comment here? 👀

packages/dbml-parse/src/core/global_modules/table/index.ts
```diff
@@ -399,6 +413,44 @@ function nodeRefereeOfInlineRef (compiler: Compiler, globalSymbol: NodeSymbol, n
   return new Report(undefined);
 }
 
+function nodeRefereeOfInlineDep (compiler: Compiler, globalSymbol: NodeSymbol, node: SyntaxNode): Report<NodeSymbol | undefined> {
+  const enclosingFa = node.parentOfKind(FunctionApplicationNode);
+  if (enclosingFa) {
+    return nodeRefereeOfInlineRef(compiler, globalSymbol, node);
+  }
+
+  if (!isExpressionAVariableNode(node)) return new Report(undefined);
```
huydo862003: Line 418 should always hold true right? is the code following (421 onwards) deadcode :?

Also, may be consider extracting the shared logic of inline ref that resolve the ref/dep operand then make nodeRefereeOfInlineDep and nodeRefereeOfInlineRef reuse this :v I don't think nodeRefereeOfInlineDep calling nodeRefereeOfINlineRef makes sense.

packages/dbml-parse/src/core/global_modules/table/interpret.ts
```diff
@@ -311,6 +311,39 @@ export class TableInterpreter {
...
+      if (result.hasValue(UNHANDLED)) return [];
+      errors.push(...result.getErrors());
+
+      const value = result.getValue() as Dep | undefined;
+      if (!value?.edges?.[0]) return [];
+
+      const prefix = (depAttr as any).value;
+      if (!(prefix instanceof PrefixExpressionNode)) return [];
+      const direction = prefix.op?.value as '->' | '<-' | undefined;
+      if (direction !== '->' && direction !== '<-') return [];
```
huydo862003: :v may be consider make `->` and `<-` check a reusable func. We already have `isRelationshipOp`. maybe add `isDependencyOp`

packages/dbml-parse/src/core/types/schemaJson/types.ts
```diff
@@ -191,6 +193,39 @@ export interface RefEndpoint {
 
 export type RelationCardinality = '1' | '*';
 
+export interface Dep {
+  schemaName: string | null;
```
huydo862003, line 201: Although `Ref` also has a `schemaName` field... I don't know if  they can have schema name in reality? 👀

packages/dbml-parse/src/core/types/schemaJson/types.ts
```diff
@@ -191,6 +193,39 @@ export interface RefEndpoint {
...
+  name: string | null;
+  edges: DepEdge[];
+  note?: {
+    value: string;
+    token: TokenPosition;
+  };
```
huydo862003, line 208: I'm wondering:

```
Dep [note: 'asdasd'] {
  table.col -> table.col [note: 'asdasd'] // can we have note here? If yes, where is it stored in this?
}
```

huydo862003: In the proposal, I see this case:

```
Dep: raw_users.email -> stg_users.email [note: 'Lowercase + validate email']
```

In this case, the note is treated in DBML as on the field, not on the Dep element :v

packages/dbml-parse/src/core/types/symbol/metadata.ts
```diff
@@ -225,6 +227,131 @@ export class RefMetadata extends NodeMetadata {
   }
 }
 
+export class DepMetadata extends NodeMetadata {
```
huydo862003, line 261: There's a problem with this:

```
// utils.dbml
Table T {
  id int
}

Table R {
  id int
}

Table E {
  id int
}

Dep {
  T.id -> R.id
  R.id -> E.id
}

// main.dbml
use {
  table T
  table R
} from './utils'

/* T.id -> R.id should be in scope, but because E is not in scope, the WHOLE dep is dropped, so nothing is in scope */
```

My current proposal:
- Add a new `DepEdgeMetadata`. `nodeMetadata` of a `Dep` field should return a `DepEdgeMetadata`.
- `owners` of `DepMetadata` should decide whether the `DepEdgeMetadata` should be included in the output file, not the whole DepMetadata.
- DepMetadata should scan its body and return a DepEdgeMetadata for each.
- When interpret DepMetadata, filter out the DepEdgeMetadata that truly belongs to this file.

huydo862003 (commented):
Please implement the language services for dep luôn nha :v can reference ref

packages/dbml-core/src/export/DbmlExporter.ts
```diff
@@ -428,13 +477,14 @@ class DbmlExporter {
 
     database.schemaIds.forEach((schemaId) => {
       const {
-        enumIds, tableIds, tableGroupIds, refIds,
-      } = model.schemas[schemaId];
+        enumIds, tableIds, tableGroupIds, refIds, depIds,
+      } = model.schemas[schemaId] as any;
```
huydo862003: why do we need `as any` here :v?

packages/dbml-core/src/export/DbmlExporter.ts
```diff
@@ -339,6 +339,55 @@ class DbmlExporter {
...
+      ? `"${ep.fieldNames[0]}"`
+      : `(${ep.fieldNames.map((f) => `"${f}"`).join(', ')})`;
+    return `${schemaPart}${tablePart}.${fieldsPart}`;
+  }
+
+  static formatDepCustomValue (value: unknown): string {
+    if (typeof value === 'string') return `'${value.replaceAll('\\', '\\\\').replaceAll('\'', '\\\'')}'`;
+    if (value === null || value === undefined) return 'null';
+    return String(value);
+  }
```
huydo862003, line 360: I think you can check `formatRecordValue`? :v or some function like `escapeString` in dbml-parse

packages/dbml-core/src/export/DbmlExporter.ts
```diff
@@ -339,6 +339,55 @@ class DbmlExporter {
...
+      const hasNote = !!dep.note;
+      const customEntries = dep.custom ? Object.entries(dep.custom) : [];
+      const hasAttrs = hasNote || customEntries.length > 0;
+
+      if (edges.length === 1 && !hasAttrs) {
+        return `Dep: ${DbmlExporter.formatDepEndpoint(edges[0].upstream)} -> ${DbmlExporter.formatDepEndpoint(edges[0].downstream)}\n`;
+      }
+
+      let str = 'Dep {\n';
+      edges.forEach((edge: any) => {
```
huydo862003: :v can we avoid any here?

packages/dbml-core/src/export/DbmlExporter.ts
```diff
@@ -339,6 +339,55 @@ class DbmlExporter {
...
+  static formatDepCustomValue (value: unknown): string {
+    if (typeof value === 'string') return `'${value.replaceAll('\\', '\\\\').replaceAll('\'', '\\\'')}'`;
+    if (value === null || value === undefined) return 'null';
+    return String(value);
+  }
+
+  static exportDeps (depIds: number[], model: NormalizedModel): string {
+    const strArr = depIds.map((depId) => {
+      const dep = (model as any).deps[depId];
+      const edges = dep.edgeIds.map((eid: number) => (model as any).depEdges[eid]);
```
huydo862003: Can we avoid any here :v

packages/dbml-core/src/parse/ANTLR/ASTGeneration/mysql/MySQLASTGen.js
```diff
@@ -9,6 +9,33 @@ import {
...
+        seen.set(key, { name: tableName, schemaName });
+      }
+    }
+    const childCount = typeof node.getChildCount === 'function' ? node.getChildCount() : 0;
+    for (let i = 0; i < childCount; i++) {
+      stack.push(node.getChild(i));
+    }
+  }
+  return Array.from(seen.values());
+}
```
huydo862003: These `extractSourceTablesFromCtx` functions are weird:
- The logic is pretty dense, hard to follow.
- Heavy regex manipulation, which is fragile.
- Highly rely on adhoc-check/runtime reflection:
  - `typeof node.getChildCount`
  - `node.constructor.name` stored in an array...

Can you think of a better way to do this?

packages/dbml-parse/src/core/types/schemaJson/types.ts
NQPhuc:
```
Dep [note: 'asdasd'] {  
  table.col -> table.col [note: 'asdasd'] // can we have note here? If yes, where is it stored in this?  
}  
```
No, we won't support dep edge level settings for now

```
In the proposal, I see this case:

Dep: raw_users.email -> stg_users.email [note: 'Lowercase + validate email']  
In this case, the note is treated in DBML as on the field, not on the Dep element :v
```
This is intentional, although the AST structure generated from that case might indicate, it's up to us to interpret it as table level setting as needed

huydo862003 (approved):

Issue #802, opened 2026-01-21
Determine the cardinality of foreign key relationships without the need for modifiers


aaditmshah-commversion:
1. **many to 1** is a non-nullable foreign key.
2. **many to [0..1]** is a nullable foreign key.
3. **[0..1] to 1** is a unique non-nullable foreign key.
4. **[0..1] to [0..1]** is a unique nullable foreign key.

I propose the following semantics.

**Note:** I shouldn't need to specify the `>` cardinality modifier. The cardinality should be inferred from the `unique`, `null`, and `not null` constraints.

```dbml
Table posts {
  // A user may create many posts. A post has only one creator.
  created_by integer [ref: users.id, not null] // many to 1

  // A user may update many posts. A post might not be updated.
  last_updated_by integer [ref: users.id, null] // many to [0..1]
}

Table compliance_training_completions {
  // An employee might not have completed the compliance training.
  // A completion record is specific to one employee.
  created_by integer [ref: employees.id, unique, not null] // [0..1] to 1
}

Table departments {
  // A department may or may not have a manager.
  // An employee can manage at most one department.
  manager_id integer [ref: employees.id, unique, null] // [0..1] to [0..1]
}
```

P.S. Also, correct me if I'm wrong. Currently, there's no way to specify the `[0..1] to 1` and `[0..1] to [0..1]` relationships in DBML.

huydo862003:
Hi @aaditmshah-commversion,

In DBML:
* To specify `[0..1] to 1` relationships, you can use the `-` type of ref, like this:
  
<img width="2878" height="1564" alt="Image" src="https://github.com/user-attachments/assets/3a70871d-6643-4a0c-a663-3d0ca0524eb4" />

It will show up as a `[0..1] to 1` as in the picture.

* The `[0..1] to [0..1]` relationship is not supported yet though.

Hope this answers your question.

aaditmshah-commversion:
<img width="1387" height="667" alt="Image" src="https://github.com/user-attachments/assets/b83ea11d-7ebd-47b7-80d5-a4a1cb1219ee" />

@hdnax It doesn't work when `a.id` is declared as a primary key.

Also, in your example the relation should be `many to [0..1]` because `b.id` is nullable and not unique. So, a row in B could be related to `[0..1]` rows in A. However, a row in A could be related to many rows in B.

[...]

PR #937, opened 2026-07-01
DBX-5320 Optional ref

huydo862003:
## Summary
* #802 
* https://app.notion.com/p/holistics/TDD-90-Zero-Optional-Relationship-Operators-317f89dc7e4980fba951ca84f2b1ff1c

### BREAKING CHANGE

All "many" endpoints are now "some" (min=1):
  - `>`: many-to-one -> some-to-one (`[1..*]-[1..1]`)
  - `<`: one-to-many -> one-to-some (`[1..1]-[1..*]`)
  - `<>`: many-to-many -> some-to-some (`[1..*]-[1..*]`)
  - `-`: one-to-one (unchanged)
  
SQL importers/db connectors will default to optional ref cardinalities & the model_structure layer in @dbml/core will adjust the cardinalities to required if it makes sense

### New features
- Introduces `?` modifier for optional (min=0) endpoints:
  - `?>`: many-to-one (`[0..*]-[1..1]`)
  - `?-?`: maybe-to-maybe (`[0..1]-[0..1]`)
  - etc.

### Changes
- 16 relationship operators (4 base x 4 optionality combinations)
  - `RefEndpoint.relation` extended from `'1' | '*'` to
  `'1'`, `'*'`, `'0..1'`, `'0..*'`, etc.
  - Lexer tokenizes `?>?`, `?<>?` etc. as single OP tokens
- SQL exporters/importers now respect nullability to generate correct relationship operators.

## Issue
(issue link here)

## Checklist

- [ ] Lint Checks Passed
- [ ] Unit Tests Passed
- [ ] Coverage Tests Passed
- [ ] Code Review


packages/dbml-parse/__tests__/utils/testHelpers.ts
```diff
@@ -112,8 +113,9 @@ function sortArray (array: unknown[]): unknown[] {
     if (typeof s === 'boolean') return 2;
     if (typeof s === 'bigint') return 3;
     if (typeof s === 'symbol') return 4;
-    if (s instanceof CompileWarning) return 5;
-    if (s instanceof CompileError) return 6;
+    if (s instanceof CompileInfo) return 5;
+    if (s instanceof CompileWarning) return 6;
+    if (s instanceof CompileError) return 7;
```
xuantho573, line 118: Should the values below be increased?

packages/dbml-parse/src/core/types/symbol/metadata.ts
```diff
@@ -145,22 +148,41 @@ export class RefMetadata extends NodeMetadata {
...
+  leftCardinality (compiler: Compiler): { min: number; max: number | '*' } | undefined {
+    const c = this.cardinalities(compiler);
+    return c ? parseCardinality(c[0]) : undefined;
+  }
+
+  rightCardinality (compiler: Compiler): { min: number; max: number | '*' } | undefined {
+    const c = this.cardinalities(compiler);
+    return c ? parseCardinality(c[1]) : undefined;
+  }
+
```
xuantho573, line 185: These 2 seem to not be used anywhere 🤔

packages/dbml-parse/src/core/types/symbol/metadata.ts
```diff
@@ -271,11 +293,30 @@ export class PartialRefMetadata extends NodeMetadata {
...
+  leftCardinality (compiler: Compiler): { min: number; max: number | '*' } | undefined {
+    const c = this.cardinalities(compiler);
+    return c ? parseCardinality(c[0]) : undefined;
+  }
+
+  rightCardinality (compiler: Compiler): { min: number; max: number | '*' } | undefined {
+    const c = this.cardinalities(compiler);
+    return c ? parseCardinality(c[1]) : undefined;
```
xuantho573: These 2 seem to not be used anywhere 🤔

packages/dbml-parse/src/core/utils/setting.ts
```diff
@@ -0,0 +1,96 @@
...
+  declaration: SyntaxNode,
+  settingName: string,
+  newValue: string,
+  source: string,
+): TextEdit | undefined {
+  const found = findSetting(source, declaration, settingName);
+  if (!found) return undefined;
+
+  return { start: found.settingNode.start, end: found.settingNode.end, newText: newValue };
+}
```
xuantho573, line 50: This seems to not be used anywhere 🤔

packages/dbml-core/__tests__/examples/parser/mssql-parse/output/with_check_add_constraint.out.json
```diff
@@ -94,7 +107,14 @@
           ],
           "relation": "1"
         }
-      ]
+      ],
+      "onDelete": null,
+      "onUpdate": null
```
xuantho573, line 112: Do we have a way to ensure either:
- Empty/null/undefined values don't appear on the output json
- They do appear
so the test output won't necessarily change in future works?

The output at `packages/dbml-core/__tests__/examples/parser/mssql-parse/output/table_constraints_indexes.out.json` at line 316 does not have `onUpdate` but I assume it'll be `onUpdate: null`.

packages/dbml-parse/__tests__/utils/testHelpers.ts
huydo862003: it's internal to the tests for consistent ordering only, wherever is fine, as long as the partial ordering among the old objects do not change :*

packages/dbml-parse/src/core/types/symbol/metadata.ts
huydo862003: yeah but just kept it though, it can be useful :pepe-cuoi-guong:

packages/dbml-parse/src/core/types/symbol/metadata.ts
huydo862003: yeah but just kept it though, it can be useful :pepe-cuoi-guong:

packages/dbml-parse/src/core/utils/setting.ts
huydo862003: it's used in #933 nhó, we're about to update fragile regexes in dbdiagram-frontend to these.

packages/dbml-core/__tests__/examples/parser/mssql-parse/output/with_check_add_constraint.out.json
huydo862003: may need to either normalize the output before outputting like in @dbml/parse :3 but maybe another pr would do

xuantho573 (approved):
🫤

packages/dbml-parse/src/core/lexer/lexer.ts
huyphung1602, line 1: On the operator() scanner (lexer/lexer.ts)

The nested if/else inside each switch case is effectively a hand-compiled trie of a fixed operator set. It works, but it's hard to eyeball for completeness: the 16 relationship forms (<, <?, ?<, ?<?, <>, <>?, ?<>, ?<>?, and the same for > and -) are spread across three levels of branching, so a missing ? branch would fail silently, and adding an operator later means threading a new path through the tree.

Since the valid operators are a small, fixed vocabulary, a maximal-munch match against an explicit set reads better and keeps the whole vocabulary in one place:
```ts
// Multi-char operators (length >= 2). Single-char ops fall through and are
// emitted as-is, so they don't need to be listed here.
const MULTI_CHAR_OPS = new Set([
  '<=', '>=', '==', '!=',            // comparison
  '-?', '?-', '?-?',                 // base `-` with an optional `?` on either side
  '>?', '?>', '?>?',                 // base `>`
  '<?', '?<', '?<?',                 // base `<`
  '<>', '<>?', '?<>', '?<>?',        // base `<>`
]);
const MAX_OP_LEN = 4; // `?<>?`

operator (c: string) {
  // `c` (1 char) is already consumed; try the longest operator first.
  for (let len = MAX_OP_LEN; len >= 2; len -= 1) {
    const candidate = this.text.substring(this.start.offset, this.start.offset + len);
    if (MULTI_CHAR_OPS.has(candidate)) {
      for (let k = 1; k < len; k += 1) this.advance();
      break;
    }
  }
  this.addToken(SyntaxTokenKind.OP);
}
```

This should be behavior-preserving:
- Longest-match handles the overlaps correctly (<>? beats <> beats <; <= beats <).
- Single-char operators and a lone ? fall through to the final addToken(OP), exactly like today.
- Adding a future operator is a one-line change to the set (bump MAX_OP_LEN if it's longer).

Not blocking — the current code is correct as far as I can tell. Just think this'll be easier to extend and to verify at a glance.

packages/dbml-parse/src/core/parser/parser.ts
huyphung1602, line 1: Two things about this table, one style and one worth confirming.

**Magic numbers.** The `left` / `right` values are Pratt binding powers (higher binds tighter, and `right = left + 1` encodes left-associativity), but nothing here says that, so every number is a small puzzle and the gaps (6, 13–15, ...) look arbitrary. Since this PR extends the table with 12 more entries, it feels like a good moment to name the levels. It also collapses the repetition: the 12 new relationship operators are all identical to `<`, currently written as twelve separate blocks.

Keeping the exact same numbers, just named and de-duplicated:

```ts
// Pratt binding powers: higher binds tighter; right = left + 1 => left-associative.
const enum Prec {
  Assign     = 2,   // =
  Equality   = 4,   // == !=
  Comparison = 7,   // < <= > >= <> and all optional-ref variants
  Additive   = 9,   // + - and the `-` optional-ref variants
  Multiplic  = 11,  // / %
  Member     = 16,  // .
}
const bp = (left: Prec) => ({ left, right: left + 1 });

const infixBindingPowerMap: Record<string, { left: number; right: number }> = {
  '.': bp(Prec.Member),
  '/': bp(Prec.Multiplic), '%': bp(Prec.Multiplic),
  '+': bp(Prec.Additive),
  '=': bp(Prec.Assign),
  '==': bp(Prec.Equality), '!=': bp(Prec.Equality),

  // one-to-one `-` and its optional-ref variants bind like `-` (additive)
  '-': bp(Prec.Additive), '-?': bp(Prec.Additive), '?-': bp(Prec.Additive), '?-?': bp(Prec.Additive),

  // every other relationship op binds like a comparison
  ...Object.fromEntries(
    ['<', '<=', '>', '>=', '<>', '?>', '>?', '?<', '<?', '?>?', '?<?', '?<>', '<>?', '?<>?']
      .map((op) => [op, bp(Prec.Comparison)]),
  ),
};
```

**Worth confirming (not just style).** The `-` family (`-`, `-?`, `?-`, `?-?`) binds at the additive level (9/10), while `<` / `>` / `<>` and their optional variants bind at the comparison level (7/8). So the one-to-one relationship binds tighter than the other relationships. I assume this is "each decorated op inherits its base op's power", but it means `-?` parses like subtraction rather than like a relationship — is that intended? Naming the levels makes this asymmetry obvious instead of hiding it in the numbers.

packages/dbml-parse/src/core/types/symbol/symbols.ts
```diff
@@ -492,26 +492,78 @@ export class ColumnSymbol extends NodeSymbol {
     return this;
   }
 
+  qualifiedName (compiler: Compiler): string {
```
huyphung1602: Can we avoid using if-in-if by return the value sooner when the tableSymbol is not existing?
```ts
qualifiedName (compiler: Compiler): string {
  const tableNode = this.declaration?.parentOfKind(ElementDeclarationNode);
  const tableSymbol = tableNode && (compiler.nodeSymbol(tableNode).getFiltered(UNHANDLED) as TableSymbol | undefined);
  if (!tableSymbol) return this.name ?? '';

  const { schema, name } = tableSymbol.interpretedName(compiler, this.filepath);
  const table = schema ? `${schema}.${name}` : name;
  return `${table}.${this.name}`;
}
```

packages/dbml-parse/src/core/types/symbol/symbols.ts
huyphung1602, line 1: `SchemaSymbol` already has a `qualifiedName` getter returning `string[]`, while this one is a method returning a dotted `string`. Same name, different shape on two classes reads confusingly. Maybe name this one `dottedName` or `qualifiedColumnName`.

packages/dbml-parse/src/core/types/symbol/symbols.ts
```diff
@@ -492,26 +492,78 @@ export class ColumnSymbol extends NodeSymbol {
...
+  isUniqueSet (compiler: Compiler): boolean {
     if (!this.declaration) return false;
     const s = compiler.nodeSettings(this.declaration).getFiltered(UNHANDLED);
 
     return !!s?.[SettingName.Unique]?.length;
   }
 
-  nullable (compiler: Compiler): boolean | undefined {
-    if (!this.declaration) return undefined;
+  unique (compiler: Compiler): boolean {
```
huyphung1602, line 534: The `if (!this.declaration) return false` is redundant: `isUniqueSet`, `pk`, and `increment` each already guard on `this.declaration`.

```ts
unique (compiler: Compiler): boolean {
  return this.isUniqueSet(compiler) || this.pk(compiler) || this.increment(compiler);
}
```

packages/dbml-parse/src/core/types/symbol/symbols.ts
```diff
@@ -492,26 +492,78 @@ export class ColumnSymbol extends NodeSymbol {
...
+        if (!idx.pk) continue;
+        if (idx.columns.some((c) => c.value === this.name)) return true;
+      }
+    }
+    return false;
+  }
+
+  // Returns whether [not null] or [null] is explicitly set
+  // true if [not null], false if [null], undefined if unspecified
+  isNotNullSet (compiler: Compiler): boolean | undefined {
```
huyphung1602, line 575: The tri-state reads clearly. Minor:

1. When a column has both `[not null]` and `[null]`, it silently returns `true` (the contradiction is not reported).
2. `compiler.nodeSettings(this.declaration).getFiltered(UNHANDLED)` is repeated across `isUniqueSet` / `nullable` / `pk` / here. A private helper would DRY it:

```ts
private settings (compiler: Compiler) {
  return this.declaration && compiler.nodeSettings(this.declaration).getFiltered(UNHANDLED);
}
```

packages/dbml-parse/src/core/types/symbol/metadata.ts
huyphung1602: Nope, please remove

packages/dbml-parse/src/core/global_modules/records/utils/constraints/fk.ts
huyphung1602, line 1: `validateForeignKey` builds each endpoint's column list like this:

```ts
const endpoint1 = table1.tableSymbol.mergedColumns(compiler)
  .filter((c) => rawEndpoint1.fieldNames.includes(c.name));
```

`filter` keeps the columns in the table's declaration order, not in the ref's `fieldNames` order. `validateEndpoint` then pairs the two sides by array position (`leftEndpoint[i]` with `rightEndpoint[i]`) and joins each side's values into one composite key in that same order. So the pairing is only correct when both tables happen to list the FK columns in the same relative order.

Example: `Ref: orders.(merchant_id, country) > merchants.(id, country_code)`. If `merchants` declares `country_code` before `id`, then `endpoint2` comes out as `[country_code, id]` while `endpoint1` is `[merchant_id, country]`, so the check compares `merchant_id` against `country_code` and `country` against `id`, producing bogus violations or missing real ones.

master ordered each side by the ref's field names, which keeps the pairing right:

```ts
const fieldSymbols = compact(rawEndpoint.fieldNames.map((name) => columnSymbolMap[name]));
```

Single-column FKs are unaffected. Fix: order each endpoint by `fieldNames` the same way.

Nit: `skipTable1 = (max1 === 1 || isOneToOne) && min2 === 0` includes `isOneToOne`, but `skipTable2 = max2 === 1 && min1 === 0` has no matching term. For operator-derived cardinalities `max` is only ever `1` or `*`, so `isOneToOne` changes nothing there; it only differs for explicit numeric ranges, where the two guards then behave asymmetrically. Make them symmetric or drop the term.

packages/dbml-parse/src/core/global_modules/ref/constraint_fixes.ts
huyphung1602, line 1: `isColumnsUnique` decides whether a set of columns is backed by a unique or primary-key constraint:

```ts
if (columns.length === 1) {
  return columns[0].unique(compiler);   // single column
}
// composite: scans the table's mergedIndexes for a matching unique/pk index
```

The single-column branch only calls `col.unique()`, which checks inline settings (`[unique]`, `[pk]`, increment). It never looks at index blocks. But the composite branch does scan `mergedIndexes`, and `ColumnSymbol.nullable()` also consults `isInPkIndex`. So a single column whose uniqueness is declared through an index (`indexes { (email) [unique] }`, or `(id) [pk]`) is treated as not-unique here.

Effect: for a single-column ref on such a column, the "one side" rule (max = 1) fires a false "should be unique or a primary key" advisory, and the "many side" rule is suppressed when it should not be. It is also inconsistent with how the composite path and `nullable()` read the same schema.

Fix: in the single-column branch, also match a single-column unique/pk index (reuse the composite branch's scan, or a shared helper like `isInPkIndex`).

Nits:

- The four rules each repeat the same ~10-line emit block (build qualified names, singular/plural message, push a `CompileInfo` on the ref node, then loop the columns pushing one per declaration). Extract an `emit(node, columns, msg, fixes)` helper; the function is ~155 lines.
- `allowOwnColFix` / `allowOtherColFix` doc comments are swapped (each describes the other side).
- Each `fixes` array starts with `opToken && suggestChangeOp(...)`, but `opToken` is already non-null after the early return at the top (`if (!op || !opToken || !cardinalities) return []`), so the `opToken &&` guard is dead.

packages/dbml-parse/src/core/types/relation.ts
huyphung1602, line 1: No bug. I traced all 16 operators through `getMultiplicities` (op to cardinality pair) and `getRelationshipOp` (pair back to op): they round-trip unchanged, and `parseCardinality` handles `*`, a bare number, and `min..max` with no off-by-one. Two cleanups:

- `normalizeCardinality` is dead code: not exported from the package barrel and no caller anywhere in the repo. Delete it, or wire it in if collapsing `N..N` to `N` was meant to run.
- `getRelationshipOp` is a ~50-line if-chain that re-enumerates the same 16 (left-bucket, right-bucket) pairs `getMultiplicities` already lists in the other direction. Two hand-written copies of the same mapping can drift. Build one from the other once (for example, invert `getMultiplicities` over `RELATIONSHIP_OPS` at load into a lookup keyed by the min/max buckets).

packages/dbml-parse/src/services/code_actions/provider.ts
```diff
@@ -0,0 +1,96 @@
...
+    const unique = uniqBy(actions, (a) => {
+      const d = a.diagnostics?.[0] as MarkerData | undefined;
+      return `${a.title}@${d?.startLineNumber}:${d?.startColumn}`;
+    });
+    if (!unique.length) return undefined;
+    return { actions: unique, dispose () {} };
+  }
+
+  // Convert a QuickFix (offset-based) to a Monaco CodeAction (line/column-based).
+  private quickFixToCodeAction (fix: QuickFix, model: TextModel, marker: MarkerData): CodeAction {
```
huyphung1602: In `quickFixToCodeAction`:

```ts
const resource = Uri.parse(fix.filepath.toUri(...)); // the edit applies to fix.filepath
const startPos = model.getPositionAt(e.start);       // but the offset is resolved against `model`
const endPos   = model.getPositionAt(e.end);
```

`e.start` / `e.end` are character offsets into `fix.filepath`'s source, but `model` is the file the user currently has open (passed in by `provideCodeActions`). When `fix.filepath` is not that open file, the offsets are converted using the wrong text, so the range points to the wrong line/column.

This is reachable. `provideCodeActions` only keeps infos on the open file (the ref), but a fix such as "mark column NOT NULL" or "add unique" edits the other endpoint's column, and that table can live in a different file. Example: the ref is in `posts.dbml` (open, so `model` is posts), but the fix edits `users.id` in `users.dbml`; its offsets get mapped through posts' text.

Effect: applying such a fix inserts the edit at the wrong spot in the target file. It only bites multi-file projects with a cross-file fix; single-file diagrams are fine.

Fix: resolve offsets against the model/source of `fix.filepath`, not `model`:

```ts
const target = monaco.editor.getModel(resource); // or the compiler's source for fix.filepath
const startPos = target.getPositionAt(e.start);
const endPos   = target.getPositionAt(e.end);
```

packages/dbml-parse/src/core/global_modules/tablePartial/interpret.ts
```diff
@@ -183,10 +183,9 @@ export class TablePartialInterpreter {
     errors.push(...typeReport.getErrors());
 
     column.pk = columnSymbol?.pk(this.compiler) || undefined;
-    column.unique = columnSymbol?.unique(this.compiler) || undefined;
+    column.unique = columnSymbol?.isUniqueSet(this.compiler) || undefined;
```
huyphung1602, line 186: The column interpret step (`table/interpret.ts:273-275`, same in `tablePartial/interpret.ts:186-188`) now writes the model from the explicit-only accessors:

```ts
column.unique   = columnSymbol?.isUniqueSet(this.compiler) || false; // was: unique(...)
column.not_null = columnSymbol?.isNotNullSet(this.compiler);         // was: !nullable(...)
```

`isUniqueSet` / `isNotNullSet` (in `symbol/symbols.ts`) only report an explicit `[unique]` / `[not null]`. The old `unique()` / `nullable()` also counted `increment` (and `pk`). The SQL exporters read `field.unique` / `field.not_null` directly (`MysqlExporter.js:79-86`, Postgres/Oracle similar) to emit `UNIQUE` / `NOT NULL`.

For a `[pk]` column this is an improvement: it no longer emits a redundant `UNIQUE NOT NULL` on top of `PRIMARY KEY`. But a column with only `[increment]` and no `[pk]`/`[unique]` used to export as `UNIQUE NOT NULL` and now exports as neither. In MySQL that leaves an `AUTO_INCREMENT` column with no key, which is invalid SQL. `[increment]` is normally paired with `[pk]`, so it is an edge case, but the old code silently covered it.

Question: is dropping the implicit `UNIQUE`/`NOT NULL` for a standalone `[increment]` intended, or should `increment` still imply them in the exported model?

packages/dbml-core/src/model_structure/database.js
```diff
@@ -243,6 +246,43 @@ class Database extends Element {
     return this.tablePartials.find((tp) => tp.name === name);
   }
 
+  // SQL importers just import everything as optional
+  // So we need to adjust the cardinality of those if they are in actuality required
+  // We only handle one-to-many/many-to-one/one-to-one here
+  adjustCardinalities () {
```
huyphung1602: The constructor calls `this.adjustCardinalities()` unconditionally (`database.js:72`). Its own comment says the SQL importers import everything as optional so it needs to adjust, but `@dbml/core` also builds a `Database` when exporting user-written DBML, so this runs on user models too.

For a ref the user deliberately made optional (`?`) whose FK column happens to be `not_null`, it upgrades the referenced side to required:

```js
const isSourceNotNull = sourceFkEndpoint.fields.every((f) => f.not_null || f.pk || f.increment);
if (isSourceNotNull) targetFkEndpoint.relation = makeCardinalityRequired(targetFkEndpoint.relation);
```

So it overrides the user's explicit cardinality. The SQL DDL likely does not change (a FK's NOT NULL comes from `field.not_null`), but the model's cardinality does, so a DBML round-trip rewrites `?>` back to `>`.

Question: should this be gated to imported models only, or is overriding a user-declared optional ref intended?

Minor: the one-to-one branch assumes `ref.endpoints[1]` is the FK source. It matches the exporter convention but relies on every importer ordering the two endpoints the same way.

packages/dbml-core/src/transform/relations.ts
```diff
@@ -0,0 +1,51 @@
...
+  CARDINALITY_MAYBE,
+  CARDINALITY_SOME,
+  CARDINALITY_MANY,
+} from '@dbml/parse';
+
+// Infer the cardinality pair for a ref from column constraints.
+// For each side:
+//   - max is determined by own column uniqueness (unique/pk -> 1, otherwise -> *)
+//   - min is determined by other column nullability (not_null -> 1, otherwise -> 0)
+export function inferMultiplicitiesFromColumns (
```
huyphung1602, line 14: `inferMultiplicitiesFromColumns` (the only export in this new file) has no caller anywhere in the repo (grepped across `packages/*/src`). Either it is intended for a downstream consumer (dbdiagram-frontend?), or it is dead code. Confirm, and drop if unused.

packages/dbml-parse/src/core/utils/setting.ts
huyphung1602: `updateSettingEdit` is exported but has no caller anywhere (constraint_fixes only imports `addSettingEdit` / `removeSettingEdit`). Dead code, drop it?

If not, please add comments?

packages/dbml-parse/src/core/utils/setting.ts
```diff
@@ -0,0 +1,96 @@
...
+} from '@/core/types/nodes';
+import type { SyntaxNode } from '@/core/types/nodes';
+import type { TextEdit } from '@/compiler/queries/transform/applyTextEdits';
+
+/**
+ * Produces a TextEdit that adds a setting to a field/element declaration.
+ * If the declaration already has a settings block [...], inserts before the closing ].
+ * Otherwise, appends a new [setting] block.
+ */
+export function addSettingEdit (declaration: SyntaxNode, setting: string): TextEdit | undefined {
```
huyphung1602, line 17: `addSettingEdit` inserts into an existing `[...]` block only for a `FunctionApplicationNode` (a field).

For an `ElementDeclarationNode` that already has an `attributeList`, it skips that branch and appends a second ` [setting]` at `declaration.end` (after the `}`), which is malformed. `findSetting` and `removeSettingEdit` both handle `ElementDeclarationNode`, so this is inconsistent. It is latent today (callers only pass a column field via `col.declaration`), but it will break if the util is reused on an element. 

Fix: read `declaration.attributeList` for the element case too, mirroring `findSetting`.

packages/dbml-core/src/transform/relations.ts
huydo862003: it's exported from @dbml/core, other consuming packages use it

packages/dbml-parse/src/core/utils/setting.ts
huydo862003: [688a61b](https://github.com/holistics/dbml/pull/937/commits/688a61b0bb23b5bab193dc57c912740915e0cf2f)

packages/dbml-parse/src/core/types/symbol/symbols.ts
huydo862003: > 1. When a column has both [not null] and [null], it silently returns true (the contradiction is not reported).

This behavior is pre-existing from dbml v1 up until now. Fixing it is a bit out of scope.

> 2. compiler.nodeSettings(this.declaration).getFiltered(UNHANDLED) is repeated across isUniqueSet / nullable / pk / here

It does not enable too much savings here.

Plus, calling `compiler.nodeSettings` is more consistent here in exchange for some DRY. Imagine we want to grep `compiler.nodeSettings` to find all major places that extract the nodeSettings, using the suggested approach would make us having to remember to grep `this.settings`.

packages/dbml-parse/src/core/types/symbol/symbols.ts
huydo862003: [44ee800](https://github.com/holistics/dbml/pull/937/commits/44ee800adf657e19445d1df2988dce5a23b2f9d7)

packages/dbml-parse/src/core/types/symbol/symbols.ts
huydo862003: [68b252d](https://github.com/holistics/dbml/pull/937/commits/68b252d06d3bae646311d018cc0df3b932ecc09b)

packages/dbml-parse/src/core/types/symbol/symbols.ts
huydo862003: [b97b85c](https://github.com/holistics/dbml/pull/937/commits/b97b85c25f946573f0138538b7a583f1f4fdcc25)

Make it return `string[]` like schema symbol

packages/dbml-parse/src/core/global_modules/records/utils/constraints/fk.ts
huydo862003: [99b7def](https://github.com/holistics/dbml/pull/937/commits/99b7def5fbd7e1c40c2e3ac8ef15f5f66bc46e6b)

> Nit: skipTable1 = (max1 === 1 || isOneToOne) && min2 === 0 includes isOneToOne, 

Actually I made a mistake there, should've been `(max1 === 1 && min2 === 0) || isOneToOne`

We skip the left table if the ref is 1-1, because the left table is the referenced side of the FK

packages/dbml-parse/src/core/global_modules/tablePartial/interpret.ts
huydo862003: https://github.com/holistics/dbml/pull/937/changes#diff-bcda920d334a953000b1f0aba7948fde02507dbdb1d93750fb83205027c3966eL502-R531

Actually, old `unique` is the new `isUniqueSet`, and i changed `unique` behavior. So the behavior is unchanged here

packages/dbml-parse/src/core/types/relation.ts
huydo862003: [37cf5fa](https://github.com/holistics/dbml/pull/937/commits/37cf5fa02a7c3e2f355d281c04037e22868f7822)

packages/dbml-parse/src/core/global_modules/ref/constraint_fixes.ts
huydo862003: [f95ff7c](https://github.com/holistics/dbml/pull/937/commits/f95ff7cd71c78cac7857b8361fa398eb9cf624bb) Fix all issues

packages/dbml-parse/src/services/code_actions/provider.ts
huydo862003: >  resolve offsets against the model/source of fix.filepath, not model

The problem with this is that dbml cant access monaco at runtime in cjs because monaco runs on the browser. -> We can't access other model and resolve position.

> Solution

In fact this isn't a problem, we already filter the fixes that are of the current "open" editor.

https://github.com/holistics/dbml/pull/937/changes#diff-991a9e137c6b2fd4ece841acad47ecedb563197b50b1a988ec93f054d032f40eR69-R71

https://github.com/holistics/dbml/pull/937/changes#diff-991a9e137c6b2fd4ece841acad47ecedb563197b50b1a988ec93f054d032f40eR32-R33

packages/dbml-core/src/model_structure/database.js
huydo862003: >  overriding a user-declared optional ref intended?

Yes it is intended:
- Users are unlikely to export dbml then import to dbml again.
- Their schemas are inconsistent, we cannot offer any guarantees as to how we will handle their schemas.

Gated imported models introduces more switch logic into the codebase but I don't think the gaining is significant.

packages/dbml-parse/src/core/lexer/lexer.ts
huydo862003: 1. This code doesn't change every much, we rarely introduce new operators.
2. Although is a bit nested, each branch is short, we can still trace pretty easily.

When this becomes a problem in the future, we can switch to a trie-based approach easily

packages/dbml-parse/src/core/parser/parser.ts
huydo862003: https://github.com/holistics/dbml/pull/937/changes/fbf14a6f02ebca9567367eaae2b1faf48a99447a

Fix nhe

huydo862003: The problem is `-` can stand for both additive and relationship.

We accept this ambiguity though, because relationship expression always have exactly one op (Ref: `a < b`).

Binding powers only have significance in expressions with multiple operators.

packages/dbml-parse/src/core/utils/setting.ts
huydo862003: `updateSettingEdit` use cases would naturally arise from `addSettingEdit` and `removeSettingEdit`, can remove it but it's very likely it will be used

packages/dbml-parse/src/core/types/symbol/metadata.ts
huydo862003: [6bf9bea](https://github.com/holistics/dbml/pull/937/commits/6bf9bea90ecc1a8a5fb321e9751fdf0242b3285a)

packages/dbml-parse/src/core/types/symbol/metadata.ts
huydo862003: [6bf9bea](https://github.com/holistics/dbml/pull/937/commits/6bf9bea90ecc1a8a5fb321e9751fdf0242b3285a)

PR #939, opened 2026-07-06
DBX-6972 Fix silent errors when an autoincrement/default PK composite component do not have a value + Refactor

huydo862003:
## Summary
* Problem: When a **composite PK column**:
  - Has increment or default constraint (so technically, `null` is fine).
  - Is omitted from records (no entry in records -> assume the default/unique values if increment)
  -> This causes `createNullErrors`/`findDuplicateErrors` to crash on undefined row values for the omitted PK columns (because the records do not have nodes for these entries)
* Fix some bugs:
  1. If PK columns are missing, fallback to report errors on the entire row, instead of crash.
  2. A PK's uniqueness check should be skipped if it has an auto-increment field.
* Refactor:
  1. Simplify signature of the records validator function which use the legacy pre-query-based compiler API.
  
Previously this would cause silent error:

```
Table T {
  id int [pk, increment]
  name string
  records (name) {
    '12' // id pk is missing, but it is increment, so the missing pk field check passes
         // however, when check if `id` is not null, we reject increment (bug) -> report not null, but there's no column `id` -> silent error
  }
}
```

## Issue
(issue link here)

## Checklist

- [ ] Lint Checks Passed
- [ ] Unit Tests Passed
- [ ] Coverage Tests Passed
- [ ] Code Review

### DIFF
--- a/packages/dbml-parse/src/core/global_modules/records/utils/constraints/unique.ts
+++ b/packages/dbml-parse/src/core/global_modules/records/utils/constraints/unique.ts
@@ -8,10 +8,8 @@ import type {
   RecordValue,
   TableRecord,
 } from '@/core/types/schemaJson';
-import { TableSymbol } from '@/core/types/symbol';
+import { TableSymbol, type ColumnSymbol } from '@/core/types/symbol';
 import {
-  type ColumnInfo,
-  columnInfoFromSymbol,
   createConstraintWarning,
   extractKeyValueWithDefault,
   formatFullColumnNames,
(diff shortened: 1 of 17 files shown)


packages/dbml-parse/src/core/global_modules/records/utils/constraints/unique.ts
```diff
@@ -8,83 +8,97 @@ import type {
...
-  });
+  return flatMap(uniqueConstraints, (uniqueColumnSymbols) =>
+    checkUniqueDuplicates(compiler, tableSymbol, uniqueColumnSymbols, rows),
+  );
 }
 
-function collectUniqueConstraints (tableSymbol: TableSymbol, columnInfos: ColumnInfo[], compiler: Compiler): string[][] {
+// Find all unique constraints in a table symbol:
+// - Inline unique
+// - Composite unique via indexes
```
xuantho573, line 36: Suggest using comments in the form `/** ... */` so the lsp can display the comments when hovering over the function

xuantho573 (approved):
🙈

PR #940, opened 2026-07-06
Fix schema-qualified TableGroup export names

### LINKED ISSUE #927: ModelExporter emits schema-qualified TableGroup names that the parser rejects (round-trip failure)
## Summary

`ModelExporter.export(model, 'dbml')` emits **schema-qualified** `TableGroup` names (e.g. `TableGroup "s"."g"`) for any table group owned by a non-default schema, but the `dbmlv2` parser **rejects** schema-qualified `TableGroup` names with *"A TableGroup name must be a single identifier"*. The exporter therefore produces DBML that its own parser cannot read back — a round-trip failure.

## Reproduction

```js
import { Parser, ModelExporter } from '@dbml/core';

// Half 1 — the parser rejects a schema-qualified TableGroup name on input:
new Parser().parse(
  'Table "s"."t" {\n  "id" int\n}\nTableGroup "s"."g" {\n  "s"."t"\n}\n',
  'dbmlv2',
); // throws: A TableGroup name must be a single identifier

// Half 2 — the exporter PRODUCES exactly that syntax for a group owned by a
// non-default schema. Build the model the way a consumer would: import postgres
// DDL whose tables live in schema "s", then attach a TableGroup to that schema
// (an ERD tool grouping tables by schema):
const model = new Parser().parse(
  'create schema "s"; create table "s"."t" ("id" int not null, primary key ("id"));',
  'postgres',
).normalize();
const schema = Object.values(model.schemas).find((s) => s.name === 's');
const gid = Math.max(0, ...Object.keys(model.tableGroups).map(Number)) + 1;
model.tableGroups[gid] = {
  id: gid, name: 'g', schemaId: schema.id, tableIds: [...schema.tableIds], note: null, color: '#cccccc',
};
schema.tableGroupIds.push(gid);
for (const tid of schema.tableIds) model.tables[tid].groupId = gid;

const exported = ModelExporter.export(model, 'dbml');
// exported contains:  TableGroup "s"."g" [color: #cccccc] { ... }

new Parser().parse(exported, 'dbmlv2');
// throws: A TableGroup name must be a single identifier
```

Output:

```
=== Half 1: parser rejects qualified TableGroup name ===
rejected: A TableGroup name must be a single identifier

=== Half 2: exporter emits qualified TableGroup name ===
exporter output: TableGroup "s"."g" [color: #cccccc] {

=== Round-trip: re-parse the exporter output ===
FAIL: A TableGroup name must be a single identifier
```

## Expected

`ModelExporter` output should always be parseable by the `dbmlv2` parser. A `TableGroup` owned by a non-default schema should either be exported with a name the parser accepts, or the parser should accept the qualified form the exporter emits.

## Actual

The exporter emits `TableGroup "<schema>"."<name>"`; the parser rejects it with *"A TableGroup name must be a single identifier"*.

## Root cause

- Exporter qualifies the group name when its schema is non-default — `packages/dbml-core/src/export/DbmlExporter.ts`, `exportTableGroups` (the `groupSchemaName` / `groupName` construction gated by `shouldPrintSchema`).
- Parser restricts the group name to a single identifier — `packages/dbml-parse/src/core/local_modules/tableGroup/validate.ts` ("A TableGroup name must be a single identifier").

The restriction appears to have been introduced by commit `f489ff7a` ("fix: allow simple name for tablegroup only") without aligning the exporter, leaving the two sides asymmetric.

## Versions

Verified on `@dbml/core` **8.2.4** and **8.2.5**; both halves are still present on `master` (exporter still qualifies, validator still rejects).

## Note: pure-DBML text does not trigger it

A group declared *unqualified* in DBML text whose members live in a named schema is parsed into the default `public` schema, so it re-exports unqualified and round-trips fine. The bug surfaces only for a consumer that builds a `NormalizedModel` with a group owned by a non-default schema (e.g. a schema-docs generator importing postgres DDL, or an ERD tool grouping tables by schema).

## Related

The `DiagramView` validator carries the same single-identifier restriction (`packages/dbml-parse/src/core/local_modules/diagramView/validate.ts`). The exporter does not emit `DiagramView` blocks, so there is no round-trip failure there today, but the parser-side limitation has the same shape.

## Workaround (for reference)

Attaching the group to the default `public` schema before export keeps the emitted name unqualified and parseable.

cyphercodes:
## Summary
* Stop DBML export from schema-qualifying `TableGroup` names so exported DBML remains parseable.
* Add a regression test covering a non-default-schema table group round trip through `ModelExporter`.

## Issue
Fixes #927

* `DbmlExporter.exportTableGroups` now emits `TableGroup "name"` while still schema-qualifying member table references when needed.

## Checklist

- [x] Lint Checks Passed
- [x] Unit Tests Passed
- [ ] Coverage Tests Passed
- [ ] Code Review

### DIFF
--- a/packages/dbml-core/__tests__/examples/model_exporter/model_exporter.spec.ts
+++ b/packages/dbml-core/__tests__/examples/model_exporter/model_exporter.spec.ts
@@ -169,6 +169,41 @@ describe('@dbml/core - DbmlExporter flags', () => {
   });
 });
 
+describe('@dbml/core - ModelExporter table groups', () => {
+  test('does not schema-qualify TableGroup names', () => {
+    const database = (new Parser()).parse(
+      'create schema "s"; create table "s"."t" ("id" int not null, primary key ("id"));',
+      'postgres',
+    );
+    const model = database.normalize();
+    const schema = Object.values(model.schemas).find((item) => item.name === 's');
+
+    if (!schema) {
+      throw new Error('Expected schema "s" to exist');
+    }
+
+    const groupId = Math.max(0, ...Object.keys(model.tableGroups).map(Number)) + 1;
+    model.tableGroups[groupId] = {
+      id: groupId,
+      name: 'g',
+      schemaId: schema.id,
+      tableIds: [...schema.tableIds],
+      note: null,
+      color: '#cccccc',
+    };
+    schema.tableGroupIds.push(groupId);
+    schema.tableIds.forEach((tableId) => {
+      model.tables[tableId].groupId = groupId;
+    });
+
+    const res = ModelExporter.export(model, 'dbml', { isNormalized: true, includeRecords: false });
+
+    expect(res).toContain('TableGroup "g" [color: #cccccc] {');
+    expect(res).not.toContain('TableGroup "s"."g"');
+    expect(() => (new Parser()).parse(res, 'dbmlv2')).not.toThrow();
+  });
+});
+
 describe('@dbml/core - ModelExporter backwards compatibility', () => {
   test('accepts boolean true as isNormalized (old signature)', () => {
     const database = (new Parser()).parse(DBML_WITH_RECORDS, 'dbmlv2');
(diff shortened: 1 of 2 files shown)


huydo862003 (approved):
Thank you for your contribution 🚀

PR #958, opened 2026-08-13
fix: resolve a dep write's block through the symbol table

NQPhuc:
## Summary

`syncDep` aimed its edits at the wrong `Dep` block, or produced none.

1. It located the block by comparing the endpoints as text.
2. It read the last part as a column only when both sides had a dot.
3. So in `Dep { raw_events.payload -> stg_events }` the left endpoint became schema `raw_events`, table `payload`, which matches nothing in the file.
4. Separately, a target naming no columns matched any edge between the same two tables.

Effect: `[color]` and `[note]` writes on a mixed-level edge produced no edits at all, and `create` appended a duplicate block each time. A table-level write landed on the pair's first column-level block instead of a direct one.

`findDepDefinition` now locates the block by resolving both endpoints to table and column symbols, the way `findRefDefinition` does for a ref. A table-level target matches only an edge with no columns on either side. When none matches, `create` writes a new block.

* Added `findDepDefinition` in `transform/utils/dep.ts`, package-private like `findRefDefinition` beside it.
* Deleted the functions that compared endpoints as text: `findDepBlocks`, `findInlineDeps`, `depBlocksFromProgram`, `inlineDepsFromProgram`, `edgeFromInfix`, `fragmentsToEndpoint`, `extractInlineDepEdges`, `endpointMatches`, `endpointsEqual`, plus the `DepBlock` and `InlineDep` types.
* `syncDep` loses its third `blocks?` parameter. `types/transform/index.d.ts` is hand-written and still declared it, so the published types now match the code.

## Type of change
- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] Updated `dbml-homepage/static/llms.txt` (if docs/features changed)
- [x] Lint Checks Passed
- [x] Unit Tests Passed (2013 in `dbml-parse`)
- [ ] Coverage Tests Passed
- [ ] Code Review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/dbml-parse/src/compiler/queries/transform/syncDep.ts
+++ b/packages/dbml-parse/src/compiler/queries/transform/syncDep.ts
@@ -14,28 +14,15 @@
  * produces a duplicate `Dep`. `remove` on an edge with no block is a no-op.
  */
 
-import { DEFAULT_ENTRY } from '@/constants';
-import Lexer from '@/core/lexer/lexer';
-import Parser from '@/core/parser/parser';
 import type { Filepath } from '@/core/types/filepath';
-import { ElementKind, SettingName } from '@/core/types/keywords';
-import { DEP_DOWNSTREAM, DEP_UPSTREAM } from '@/core/types/schemaJson';
-import {
-  ElementDeclarationNode,
-  FunctionApplicationNode,
-  InfixExpressionNode,
-  ListExpressionNode,
-  PrefixExpressionNode,
-  ProgramNode,
-  SyntaxNodeIdGenerator,
-} from '@/core/types/nodes';
-import { destructureComplexVariable, extractSettingName } from '@/core/utils/expression';
+import { SettingName } from '@/core/types/keywords';
+import { ElementDeclarationNode } from '@/core/types/nodes';
 import type Compiler from '../../index';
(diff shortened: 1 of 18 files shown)


packages/dbml-parse/src/compiler/queries/transform/syncDep.ts
huydo862003, line 1: There are some functions that don't seem to be used anywhere:
- findInlineDeps: unused + unexported from dbml/parse -> Can we delete this 🤔 
- findDepBlocks: unused but exported from dbml/parse -> If not used anywhere else, we can also delete this.
- depBlocksFromProgram
- inlineDepsFromProgram
- DepBlock and InlineDep type

:v I dont have much context. But if we're not intending to use these anywhere, we can delete these 🤔

huydo862003 (approved):

packages/dbml-parse/src/compiler/queries/transform/syncDep.ts
NQPhuc: removed un-unused (and incorrect) code in https://github.com/holistics/dbml/pull/958/commits/f02ceb2313e122402567337609bdb0f2c9ce5edc

PR #962, opened 2026-09-01
Correct semantics for enums and nulls

hlieberman:
## Summary

Currently, DBML has an incorrect semantic behavior when specifying exporting an enum in MSSQL and Oracle. For both of these database formats, the exporter unconditionally sets the column to be NOT NULL. At first, this seems to intrinsically make sense --- an enum is a closed set of specified values, so what would NULL even mean.

While this is logical, it is not the way NULL is implemented in any of the major databases. Both PostgreSQL and MySQL implement enums directly, and MySQL is explicit about the fact that NULL is a supported value (https://dev.mysql.com/doc/refman/8.4/en/enum.html#enum-nulls).  PostgreSQL's docs don't contain a similar assertion, but it also supports NULL in columns with enums.

This change does cause the exporter to potentially emit different SQL than it would have previously, but it ensures that all the databases behave the same based on the same input.  It also fixes a bug that occurs when you specify a column to be both an enum and NOT NULL in the MSSQL/Oracle exporter.

## Checklist

- [X] Documentation (if necessary)
- [X] Updated `dbml-homepage/static/llms.txt` (if docs/features changed)
- [X] Lint Checks Passed
- [X] Unit Tests Passed
- [X] Coverage Tests Passed
- [X] Integration Tests Passed
- [ ] Code Review

### DIFF
--- a/packages/dbml-cli/__tests__/dbml2sql/filename --oracle --out-file/expect-out-files/schema.sql
+++ b/packages/dbml-cli/__tests__/dbml2sql/filename --oracle --out-file/expect-out-files/schema.sql
@@ -1,7 +1,7 @@
 CREATE TABLE "orders" (
   "id" int GENERATED AS IDENTITY PRIMARY KEY,
   "user_id" int UNIQUE NOT NULL,
-  "status" nvarchar2(255) NOT NULL CHECK ("status" IN ('created', 'running', 'done', 'failure'))
+  "status" nvarchar2(255) CHECK ("status" IN ('created', 'running', 'done', 'failure'))
 );
 
 CREATE TABLE "order_items" (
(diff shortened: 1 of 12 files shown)


huydo862003 (dismissed):

huydo862003:
Hi @hlieberman, thanks for your contribution.

Please note that for your change to be reflected on dbdiagram, you will have to wait for a few weeks. Thank you again!

huydo862003:
Hi @hlieberman, can you help fix @dbml-cli tests since they are broken from your changes?

hlieberman:
My apologies; I didn't think there would be end-to-end tests in the @dbml/cli area, so I only ran the ones in @dbml/core.  Silly mistake!

hlieberman:
Lint failures are unrelated to the PR.  Same with the rule_check.

hlieberman:
Embarrassing.  I added fixes, and then ran the same command as before... skipping the CLI tests.  Repushed.

huydo862003 (approved):

